PR #2, opened 2023-01-18
Logger manager

Pennedrose:
## Overview

This PR will implement the logging manager that makes it much simpler to log data. Automatically starts logging motor temp, current, and RPM. 

## Design Doc
 Anthony said design doc was not needed. 

## Tests Ran
 Built the code and tested that the values given in AdvantageScope were accurate.


AnthonyFurman:
I think the motor logging should have a separate folder under the logs.
Also, don't have an enter between the javadoc comment and the start of the function.

Otherwise looks good.

AnthonyFurman (approved):

PR #3, opened 2023-01-30
Unittest

GLRoylance:
## Overview

This PR adds JUnit tests. It also adds a trivial unit test to show the team how to add such tests.

## Design Doc

The Master Software Tests Document should be updated to include unit tests. These tests are automated. They can include robot simulations without using Driver Station.

For example, I expect the team to make unit tests that check April Tag code. If BeeLink gives x, y, theta for M1, is the updated robot pose correct?

## Tests Ran

I ran build robot code with no errors. The build successfully ran the unit test.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -65,6 +65,8 @@ dependencies {
     nativeRelease wpi.java.deps.wpilibJniRelease(wpi.platforms.desktop)
     nativeRelease wpi.java.vendor.jniRelease(wpi.platforms.desktop)
     simulationRelease wpi.sim.enableRelease()
+
+    testImplementation 'junit:junit:4.12'
 }
 
 // Simulation configuration (e.g. environment variables).
--- a/build.gradle
+++ b/build.gradle
@@ -66,7 +66,18 @@ dependencies {
     nativeRelease wpi.java.vendor.jniRelease(wpi.platforms.desktop)
     simulationRelease wpi.sim.enableRelease()
 
-    testImplementation 'junit:junit:4.12'
+    // JUnit 4
+    // testImplementation 'junit:junit:4.13'
+
+    // JUnit 5. WPILIB wants to enable JUnit 5 by default
+    testImplementation("org.junit.jupiter:junit-jupiter-api:5.6.2")
+    testImplementation("org.junit.jupiter:junit-jupiter-params:5.6.2")
+    testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:5.6.2")
+}
+
+test {
+    useJUnitPlatform()
+    systemProperty 'junit.jupiter.extensions.autodetection.enabled', 'true'
 }
 
 // Simulation configuration (e.g. environment variables).
--- a/src/test/java/frc/robot/util/ArithTest.java
+++ b/src/test/java/frc/robot/util/ArithTest.java
@@ -1,21 +1,42 @@
 package frc.robot.util;
 
-import static org.junit.Assert.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 
-import org.junit.*;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
 
 /**
  * Example of a JUnit test class.
  * This test should run everytime someone builds the robot code.
+ * See https://docs.wpilib.org/en/stable/docs/software/wpilib-tools/robot-simulation/unit-testing.html
+ * 
+ * To disable a test, annotate with Disabled
  */
 public class ArithTest {
+
+    @BeforeEach
+    public void prepare() {}
+
+    @AfterEach
+    public void cleanup() {}
 
     /**
      * Test if floating point addition works.
      */
     @Test
     public void testSimpleArith() {
-        assertEquals("addition does not work", 5.0, 2.0 + 3.0,  0.0001);
+        assertEquals(5.0, 2.0 + 3.0,  0.0001);
+    }
+
+    /**
+     * Here is a disabled test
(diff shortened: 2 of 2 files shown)


AnthonyFurman:
Last year we removed unit tests because we found them counterproductive and they would cause issues. They were also not used well, basically just given to students as busywork, rather than writing productive tests. That's partially a problem with how we did programming last year but also a problem with the implementation and the ability of the tests. Additionally we would have to rewrite subsystem initialization to allow mocking of the components which would obfuscate the code. So while I initially had them in the 2023 code structure, I later removed them.

However, this year it seems we are aiming to have more mathematically complicated code. Mathematical checks can be run through unit testing so for that purpose I support it. The other thing is simulator unit tests which we did not do last year and still needs to be tested by a student. That would be a very useful way to use unit tests.
So I think we can add them.

But, I'm not sure the unit tests need to run every build. Especially for the simulator unit tests, starting the simulator environment and running the test every build would mean a very significant development slowdown.
I think a student could look into use of the simulator in unit tests, online I found no references of it being done but I don't see why it can't be done. I'll see who can be assigned to that work. Ideally it would be fully programmed by a student, I thought there was a policy against mentors committing to main, and it would be very simple replicate what this commit does.

So this PR might be closed, but I will prioritize a student getting this done.

GLRoylance:
Anthony,

It does not take long to run the unit tests at every build. Even the
simulations run quickly. Of course, unit tests should not simulate
autoroutines in real time....

Last year's code did not need to mock (Mockito) devices because the devices
were already mocked.

WPILIB docs describe simulations. Those simulations can be used in both
SimGUI and unit tests.

I take policy as guidelines. The substantial point is that mentors should
not be writing software; that is something that the students should be
doing. However, if students have written some code, then I see no problem
with showing how that code could be improved.

Here are some questions for you.

Why doesn't the team have a RobotContainer? The team is not following the
examples provided by WPI. Why?

Why are there so many constants classes? When should declarations be in a
separate class and when can they be included in a subsystem or other class?

Why does Constants have declarations such as

  public static final OIConstants oi = new OIConstants();
  public static final DriveConstants drive = new DriveConstants();
  public static final AutoConstants auto = new AutoConstants();
 ?

Why aren't the declarations in, for example, OIConstants and DriveConstants
declared public static final?

Why do you need a DoNothing command?

Why does the team use so many globals? For example, Driver accesses
Robot.drive globally rather than taking drive as an argument. Globals are
messy.

Jerry

AnthonyFurman:
I'm not sure that the simulations run that quickly. I guess the simulation could be run faster than real time, but I think we should test this first. When I talked to Vadim about this he said it was probably a bad idea to try to run the simulator every single build.

I'm not sure what you mean about the mocking because last year we certainly did mock the devices. Or do you mean that mocking was unnecessary?

Ok, I agree with the next two paragraphs.

Robot Container only served as a class that Robot made an instance of that held a few things (basically just the subsystems). Richie moved the subsystems to Robot.java and deleted robot container and we never looked back. RobotContainer is not terrible but there's really no need for another class with an instance just to hold a few variables. Otherwise nearly everything is by the WPI books,

There are a lot of Constants classes because there are a lot of constants :). WPI standard is to have them in one file which is hard to look through and troublesome. 
About the declarations, maybe you can clarify what you mean by declarations? Which kind? Declarations of constants is always in a constants file.

Constants makes an instance of each constants file that so that you can access the specific constants by going Constants.drive or Constants.oi, etc. Personally I find it a little weird too but it's convenient to write. It could be changed.

Richie added the DoNothing command, I think it was for ConditionalCommands when he just wanted nothing to happen if the condition failed (and for it to be clear in the code), though this can also be done with the unless() decorator or maybe making custom command composition. So it could be changed.

The only such static public variables are the subsystems. It's much easier to handle them like this than passing the reference around to every single subsystem. I suppose it could be messy due to access issues but methods and fields in the subsystem itself are almost always private unless they are safe to be modified.

GLRoylance:
HI Anthony,

I reworked a lot of last year's code, and that included restoring and
improving some simulations.  One of the simulations was bringing the
shooter wheel up to speed using a realtime simulation. The unit tests are
run in parallel.

The ;Mockito mocking was not necessary. The mocks already existed.

It's more complicated. RobotContainer was deleted, but a lot of its
functionality was moved to ShuffleboardManager. ShuffleboardManager should
not know so much about the robot. The Shuffleboard is a service that other
classes should use; it should not be managing the robot. It is quite
possible that ShuffleboardManager killed us at Chezy Champs.

The WPI books say something much different. Most commands were accessing
subsystems through globals.

I'm using declaration as the definition of a constant. To me, the only
reason to have constants declared in separate classes is when those classes
will be used by several other classes or there is an allocation issue.
Consider a robot arm subsystem. The motor currents and PID constants are
local to the subsystem and should be declared in the subsystem class rather
than in a separate constants class. However, I would put the motor port
assignments in a shared constants file. That way, I can see all of the used
motor ports and can easily choose an unused port for a new motor.

The current constant system requires unneeded indirection. To access an OI
constant, the reference logically goes through the oi instance. If the
OIConstants file is imported (or OI is a subclass in Constants), then the
value can be static final  and use a compiled literal. The syntax is
similar:

Constants.oi.kDriverJoy

versus

OIConstants.kDriverJoy

More significantly, using the latter allows constants to follow the static
final pattern.

(kDriverJoy is also an example of a constant that is not used by any other
class; it can be local to the Driver class.)

The DoNothing and many BooleanSupplier were because the team wrote Command
code to evaluate conditionals during execution when they could have been
evaluated during instantiation. The team made a code more complicated than
it needed to be.

WPI considers using globals for subsystems bad practice. Passing subsystems
is verbose, but it adds clarity. It was awkward last year because there
would be a left and right arm and a left and right extender.

AnthonyFurman (approved):

PR #4, opened 2023-01-30
Change code so SimGUI will run

GLRoylance:
## Overview

This PR will allow SimGUI to run.

It would not run because the motor controllers had negative port numbers and the simulator returned a bad firmware revision.

## Tests Ran

Ran the simulator.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -160,6 +160,9 @@ bin/
 
 # Simulation GUI and other tools window save file
 *-window.json
+simgui-ds.json
+simgui.json
+*.wpilog
 
 # File that has git and version data
 src/main/java/frc/robot/util/BuildData.java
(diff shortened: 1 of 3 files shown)


AnthonyFurman:
It's probably better to still check the firmware. Also, the simgui file is useful so you don't have to setup your keyboard every time. I added some of these in #6

GLRoylance:
You will not have to set up your SimGUI each time; the file will stick
around on one's local branch. If the json files are part of the remote
branch, then everyone gets the same json assignments.

Last year's simulation was uneven. Sometimes the firmware value would be
returned, and sometimes it would return -1. I think it was a timing issue:
the simulator would wait for simulated CAN messages (i.e, the simulator
would ask the motor its firmware version).

AnthonyFurman:
Added in the other PR

PR #5, opened 2023-01-30
Testmode

GLRoylance:
## Overview

This PR deprecates Robot.isTestMode() because RobotBase.isTest() already provides that functionality.

## Tests Ran

Cannot run SimGUI for reasons in another PR.


src/main/java/frc/robot/Robot.java
```diff
@@ -26,7 +26,8 @@ public class Robot extends TimedRobot {
   public static ShuffleboardManager shuffleboard;
   public static Drivetrain drive;
 
-  private static boolean isTestMode = false;
+  // this should be a RobotContainer
```
AnthonyFurman, line 29: Sorry, I'm not sure what this comment means or what this variable is for?

src/main/java/frc/robot/Robot.java
```diff
@@ -135,7 +131,15 @@ public Command getAutonomousCommand() {
...
   }
 
-  public static boolean isTestMode() {
-    return isTestMode;
+  /**
+   * Whether we are currently in Test Mode.
+   * @deprecated use RobotBase.isTest()
+   * Also confused about the Robot being STATIC.
+   * Main builds a single INSTANCE of the robot.
+   * At a later time, it will call the robotInit() method.
```
AnthonyFurman, line 139: That's true. Honestly at this stage in development I don't think we need to deprecate the method we can just change it. Probably at any stage of development as we can coordinate with the rest of the team. I also think including comments like that are not a good idea on the main branch, the main branch should be more or less finalized. And when hovering over a function the javadoc should describe the function, not thoughts about the implementation of robot.java. 
As for the static/instance issue, I can add a getInstance() method which can be used by Main.java and for getting the test mode. The robot class is a singleton.

GLRoylance:
The original code created isTestMode(), but that functionality is already
available in RobotBase.isTest().

The code makes several references to Robot static variables, but
RobotBase.isTest() is not a static method. There needs to be a route for to
access the robot instance so the isTest() method can be accessed.

The RobotContainer comment is much more involved. The team is not following
WPI advice.

Jerry

GLRoylance:
Hi Anthony,

I deprecated the method because I do not know whether other branches are
using the method. If isTestMode() is deleted, then it would break the
branches that call it. With a deprecated method, the branches will still
work when they merge main. Some warnings will tell them to change the calls
to isTestMode().

Comments on a deprecated function? Nobody should be referencing the method,
so the comments will not ordinarily display. Eventually, the deprecated
methods are deleted. When we know that others are not using the method, the
method can be deleted.

Yes, the Robot class is sort of a singleton; in real life there is only one
motor on port 13, and it cannot be shared. Even simulation has limited
resources. But the Robot constructor is not private, so it logically does
not prevent the construction of multiple instances.

Main.java can be left alone.

Managing globals is complicated. The basic goal is avoid a lot of globals.
WPI wants RobotContainer to manage access to all of the subsystems. If
RobotContainer does not give a method explicit access to a subsystem, then
that method cannot access the subsystem.

Jerry

AnthonyFurman:
Yeah, I understand the deprecation will keep their code from breaking, but at this point there is only one branch that uses it and it can quickly be changed. We have less than 13 programmers, I think we can just let them know about changes, at least for now.

I think what I mean by the comment above the method is that the comment does not particularly relate to the function but is instead a comment about your thoughts on the code structure. That shouldn't be the comment that prefaces a function. The thoughts in that comment are also what you're saying in this PR (or really across 2 PRs) which is better but still not ideal. We should probably discuss these issues in person or over email rather than javadoc comments.

About a singleton and Main.java. I think I wasn't clear. We use Robot.java effectively as a singleton and I can make it one by adding a getInstance() method. Then the constructor can be made private and Main.java can use getInstance(). I've already programmed that and it worked enough to run the simulator. If you see any issues that could come from that let me know but it would be more consistent with how we use Robot.java anyway.

I think I already commented about globals and robot container in the other PR. 

Maybe we should talk about this more in person.

AnthonyFurman:
I added an example of the singleton changes in #6 so it wouldn't affect your other changes.

AnthonyFurman:
After our conversation today, we've decided to implement this with robotcontainer

PR #6, opened 2023-01-30
Sim working and cleanup

AnthonyFurman:
Simulator ran and appears to be working.

Mostly a consolidation of @GLRoylance 's changes in the other PRs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -160,6 +160,10 @@ bin/
 
 # Simulation GUI and other tools window save file
 *-window.json
+*simgui.json
+
+# Simulation puts log files in root of repo
+*.wpilog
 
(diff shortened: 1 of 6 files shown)


GLRoylance:
Down vote changes.

Do not add code to Main.java.

Do not add getInstance() to Robot. It is not a strict singleton.

AnthonyFurman:
Jerry, I reverted the singleton change, next time we meet we can talk about how we handle robot.java and all the subsystems.

I kept simgui-ds, but it is gitignored, so you don't have to set up the controls if you use a new computer or someone is running simulation the first time.

PandaNinja10 (approved):
no comments of worth.

PR #9, opened 2023-02-09
Code structure changes

AnthonyFurman:
## Overview

This PR will add RobotContainer.java back, remove ShuffleboardManager, change constants to be public static, and force to be subsystems passed through parameters rather than as global statics.

## Tests Ran
The code builds.


GLRoylance:
Looks good to me.

PandaNinja10 (commented):
the test types were removed, then not reimplemented is this a permenent change and if so why?

AnthonyFurman:
Yes, instead of constantly checking the test type, it is easier to add the test command to shuffleboard, which will create a button that schedules the command. 
Also there will be the test controller on port 2, so you can switch the controller port in DS so if a test requires buttons it can easily be run like that.

PandaNinja10 (approved):

PR #10, opened 2023-02-11
4 bar arm

MagicEmu:
## Overview

Four bar arm control code, no constants

## Design Doc

https://docs.google.com/document/d/1BCxfgVRcpiqJ0dqkICla7OOI6J3N9_fGweoOEehTh8E/edit
Implemented four bar arm control using PID

## Tests Ran

tests to be implemented with simulator


src/main/java/frc/robot/commands/arm/ExtendToPosition.java
```diff
@@ -0,0 +1,34 @@
...
+    m_arm.setEncoderPosition(0);
+  }
+  @Override
+  public void execute() {
+    m_arm.setMotor(m_pid.calculate(m_arm.getEncoderPosition(), m_armSetpoint));
+  }
+  @Override
+  public boolean isFinished() {
+    double error = m_armSetpoint - m_arm.getEncoderPosition();
+    return error <= -1 || error >= -1; // error values TBD
```
vfurman-gh, line 28: Use Math.abs(error)? Also, introduce a constant kMaxFinalError, so it will look like `return Math.abs(error) < MaxFinalError`

src/main/java/frc/robot/commands/arm/Retract.java
```diff
@@ -0,0 +1,32 @@
+package frc.robot.commands.arm;
+
+import edu.wpi.first.math.controller.PIDController;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.constants.ArmConstants;
+import frc.robot.subsystems.FourBarArm;
+
+public class Retract extends CommandBase {
```
vfurman-gh, line 8: Add a comment on how is this command different from Extend command?

src/main/java/frc/robot/commands/arm/Retract.java
```diff
@@ -0,0 +1,32 @@
...
+  FourBarArm m_arm;
+  PIDController m_pid;
+  public Retract(FourBarArm arm, PIDController pid) {
```
vfurman-gh, line 11: Ca this class be implemented in terms of Extend Command?

GLRoylance:
This code is mixing the commands and the subsystem. Command should start things moving. The subsystem should control that movement.

The commands need to declare that they use the subsystem. See addRequirements(m_arm).

The PID controller should not be passed into the Commands. The PID controller should only be used by the subsystem. The subsystem abstracts the needed motor control. Command should not worry about those details.

The commands should start things moving by calling methods in FourBarArm.

Command.initialize should set the setpoint with something such as m_arm.setTargetPosition(setpoint) rather than using the PID controller. Notice that Retract(arm) is ExtendToPosition(arm, 0)

Command.execute should not be running the PID loop. The subsystem should update the control loop in its .periodic() method.

Command.isFinished() should call m_arm.isFinished(). That method should use m_PID.atSetpoint(). The tolerance is set with m_PID.setTolerance(). That tolerance should be known to the subsystem; the Commands need not know it.

Commands should not be reseting encoders.

src/main/java/frc/robot/commands/arm/ExtendToPosition.java
```diff
@@ -0,0 +1,34 @@
...
+
+import edu.wpi.first.math.controller.PIDController;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.subsystems.FourBarArm;
+
+public class ExtendToPosition extends CommandBase {
+  FourBarArm m_arm;
+  PIDController m_pid;
+  double m_armSetpoint;
+  public ExtendToPosition(FourBarArm arm, PIDController pid, double setpoint) {
```
AnthonyFurman, line 11: Don't pass in a PID, you can make one in the subsystem

src/main/java/frc/robot/constants/Constants.java
```diff
@@ -19,4 +19,6 @@ public final class Constants {
   public static final AutoConstants auto = new AutoConstants();
   @Deprecated
   public static final FalconConstants falcon = new FalconConstants();
+  @Deprecated
+  public static final ArmConstants arm = new ArmConstants();
```
AnthonyFurman, line 23: Don't need this

src/main/java/frc/robot/subsystems/FourBarArm.java
```diff
@@ -0,0 +1,41 @@
...
+    return m_encoder.getPosition();
+  }
+  public void setEncoderPosition(double position) {
+    m_encoder.setPosition(position);
+  }
+
+  public void setMotor(double val) {
+    m_motor.set(val);
+  }
+  public double getMotorValue() {
```
AnthonyFurman, line 34: getMotor() or getMotorPower()

src/main/java/frc/robot/subsystems/FourBarArm.java
```diff
@@ -0,0 +1,41 @@
...
+  private final PIDController m_pid;
+  private final RelativeEncoder m_encoder;
+
+  public FourBarArm() {
+    m_motor = new CANSparkMax(ArmConstants.motorID, MotorType.kBrushless);
+    m_motor.setIdleMode(IdleMode.kBrake);
+    m_encoder = m_motor.getAlternateEncoder(SparkMaxAlternateEncoder.Type.kQuadrature, 8192);
+    m_pid = new PIDController(ArmConstants.kP, ArmConstants.kI, ArmConstants.kD);
+  }
+
```
AnthonyFurman, line 23: Add an @Override periodic method to run the pid if the subsystem is enabled here

src/main/java/frc/robot/commands/arm/Retract.java
GLRoylance: Yes. Retract = new ExtendToPosition(arm, 0). It could be made as a sequential command. Alternatively, do not use a Retract command.

src/main/java/frc/robot/commands/arm/ExtendToPosition.java
```diff
@@ -0,0 +1,34 @@
...
+    m_arm = arm;
+    m_pid = pid; 
+    m_armSetpoint = setpoint;
+  }
+  @Override
+  public void initialize() {
+    m_pid.reset();
+    m_arm.setEncoderPosition(0);
```
AnthonyFurman, line 19: Don't reset the encoders here, otherwise it won't work if the arm is in a different angle. I would that run in the init of the subsystem.

src/main/java/frc/robot/subsystems/FourBarArm.java
```diff
@@ -0,0 +1,41 @@
...
+  public double getEncoderPosition() {
+    return m_encoder.getPosition();
```
AnthonyFurman, line 25: Look into scaling the encoder value

src/main/java/frc/robot/commands/arm/ExtendToPosition.java
vfurman-gh: Based on Jerry's comment below, this is no longer needed.

AnthonyFurman (changes requested):
I think the only thing that should be changed is in Operator.java, the Y button is binded to two different actions, and the arm shouldn't retract immediately after it extends. We can add that part after we have a working intake. So for now it should just extend.

@GLRoylance @richtan @PandaNinja10 other than that small change looks good?

AnthonyFurman:
Also, can you update the design doc, I think the low-level design section is out of date.

src/main/java/frc/robot/subsystems/FourBarArm.java
```diff
@@ -0,0 +1,48 @@
...
+  @Override
+  public void periodic() {
+    m_motor.set(MathUtil.clamp(m_pid.calculate(m_encoder.getPosition()), ArmConstants.minMotorPower, ArmConstants.maxMotorPower));
+  }
+
+  public boolean atSetpoint() {
+    return m_pid.atSetpoint();
+  }
+
+  public void end() {
```
AnthonyFurman, line 45: maybe just call this stopMotors()

richtan:
Where is the code for the absolute encoder?

GLRoylance:
m_arm.end() is pointless: The m_motor.set() will be overridden at the next
periodic().

Once the PID has reached its setpoint, then the goal of the command has
been achieved. That does not mean we should turn off the motor (set the
voltage to zero). Instead, we should hold the position until we are
commanded to a new position. Consider the sequence

Move FourBar to the middle position.
Drop Cone.

We do not want the arm to move while we are dropping the cone. We want the
PID to run even after the commanded move has finished.

Also, brake mode does not apply brakes or hold the current position.  It
shorts the motor winding. That increases drag, but the drag force is
proportional to velocity. At low velocities, the drag force is small.

I would move the motor id constant to the main constants file (and assign
all motor ids there). That way we can see all of the motor id assignments
in one place.

I would move all of the other constants to the subsystem FourBar.java. The
subsystem is the only class that is interested in the PID values. The
subsystem is also the place where all the position values are made
relevant. There is little difference between obtaining static final values
from constants.ArmConstants or from FourBar; none of the non-static values
are accessible from FourBar (they must be accessed from an instance).

I would make the positions an enum.

Always assign a valid CAN id (not -1). The value may change later, but a -1
will kill SimGUI. The main branch should always run SimGUI.

AnthonyFurman:
Gerry, about the constants, that would be different than the rest of the code structure, so I'm not sure it should be done that way. The rest I agree with

GLRoylance:
I believe in private, and I believe in locality.

If a constant is only used in one class, why should it be advertised to the
public?

If I'm working on code for one mechanism, why do I want to spread that code
over two classes?

vfurman-gh:
On Mon, Feb 13, 2023, 12:03 PM GLRoylance ***@***.***> wrote:

> I believe in private, and I believe in locality.
>
> If a constant is only used in one class, why should it be advertised to the
> public?
>
There is a value in having all constants in one place, so they can be
found, examined and tweaked quickly.

AnthonyFurman:
What vadim said.
Any specific numbers (ie. not 0, 1, 2) should be stored as a variable in a separate location. they could be stored at the top of the file, but that creates a long list of values you have to scroll past, that most of the time you are not modifying anyway. For ease of use they are in a separate file, all the students are very familiar with using the constants and modifying the values. I want to stress that last point, it is easier to use for all of the students, even (and especially) new students. It would be cumbersome to look through the subsystem for those values, or worse, look through both the subsystem file and the constants file.

GLRoylance:
I disagree with the last two comments.

If constants are shared among several classes, then a constants file makes sense. If the constants are used only in  a class, then put them in that class. They can be found quickly without opening a second file.

The robotics class is not using one constants class (where a text search can quickly find a named constant). Instead it uses several constants files, so there may be several files to check. Usually the appropriate constants file matches the subsystem (but see FourBarArm and ArmConstants). Yes, most references will include the class name (e.g., ArmConstants.kP).

There are not long lists of constants for a subsystem.  The ArmConstants class is only 31 lines long. I would put the constant definitions next to relevant classes. For example, the PID coefficients would be declared next to the PIDController declaration (where they can be used immediately). The arm's physical parameters can be located next to the SingleJointedArmSim declaration. If the PID isn't working, coders would be looking at the FourBarArm class. If the simulation isn't working, coders sould be looking at the FourBarArm class.

Finding a definition is easy. In Visual Code, go to where the definition/constant is used (e.g., where the PID instance is created using kP, kI, and kD), right click, select find all implementations, and then click the found implementation.

src/main/java/frc/robot/constants/ArmConstants.java
```diff
@@ -0,0 +1,30 @@
...
+  public static final double intakePosition = -1;
+  public static final double middlePosiiton = -1;
+  public static final double topPosition = -1;
+
+  // simulation constants
+  public static final DCMotor armSimMotor = DCMotor.getNEO(1);
+  public static final double armReduction = 45;
+  public static final double armMass = 2.05;
+  public static final double armLength = .24;
+  public static final double armMOI = .1194;
```
PandaNinja10, line 29: what is  MOI

src/main/java/frc/robot/controls/Operator.java
```diff
@@ -11,7 +13,10 @@ public class Operator {
   /**
    * Configures all of the operator controls.
    */
-  public static void configureControls() {
-    operator.get(Button.A).onTrue(new DoNothing());
+  public static void configureControls(FourBarArm arm) {
+    operator.get(Button.Y).onTrue(new ExtendToPosition(arm, ArmConstants.topPosition));
+    operator.get(Button.X).onTrue(new ExtendToPosition(arm, ArmConstants.middlePosiiton));
+    operator.get(Button.A).onTrue(new ExtendToPosition(arm, ArmConstants.intakePosition));
+    operator.get(Button.Y).onTrue(new ExtendToPosition(arm, ArmConstants.shelfPosition));
```
PandaNinja10, line 20: Y is bound to 2 positions

GLRoylance:
Moment of Inertia.

It is like mass in F = ma except it is for angular acceleration.

AnthonyFurman (approved):

PR #12, opened 2023-02-11
read auto routine in autonomousInit

GLRoylance:
## Overview

This PR reads the autonomous Command during autonomousInit().

## Tests Ran

Simulator and checked NetworkTables

## Additional Notes

It was done differently last year because reading the autonomous command did a lot of busy work.

### DIFF
--- a/src/main/java/frc/robot/Robot.java
+++ b/src/main/java/frc/robot/Robot.java
@@ -54,13 +54,17 @@ public class Robot extends TimedRobot {
 
   /** This function is called periodically when the robot is disabled */
   @Override
-  public void disabledPeriodic() {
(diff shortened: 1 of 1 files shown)


AnthonyFurman:
both this year and last year, it seems like this is the getAutonomousCommand function:
```
}
```
I know there was some other stuff that would set up shuffleboard if that's what you mean?

GLRoylance:
Yes. Just fetching the selected command is fast, but setting up
Shuffleboard was slow.

Also, I doubt the tactic of doing the work while exiting Disabled instead
of entering Autonomous had much impact. The workload would still overflow
because there was more than 20 ms worth of work to do.

AnthonyFurman (approved):

PR #13, opened 2023-02-11
Deploying bar

zhoutony05:
## Overview
This PR is a preliminary write-up of the deployingbar code without constants. There may be changes, as Allan is currently redesigning the bar.

## Design Doc
(https://docs.google.com/document/d/1BCxfgVRcpiqJ0dqkICla7OOI6J3N9_fGweoOEehTh8E/edit?usp=sharing)


AnthonyFurman:
You need to update your branch from main, there are some significant code structure changes. Primarily, the instance of the subsystem will be a private variable in robot container, and the constants are public static, accessed like DeployingBarConstants.kExampleConstant.

zhoutony05:
I made some of the changes you suggested. Let me know if there are any others.

AnthonyFurman:
Ok, next time update the branch from main. If you use the github desktop gui then it's the branch button at the top then update from main. I've done that for you.

src/main/java/frc/robot/commands/ExtendDeployingBar.java
```diff
@@ -0,0 +1,48 @@
+package frc.robot.commands;
+
+import frc.robot.subsystems.DeployingBar;
+
+import edu.wpi.first.wpilibj2.command.CommandBase;
+
+public class ExtendDeployingBar extends CommandBase {
+  private final DeployingBar m_deployingbar;
+  double setpoint;
```
AnthonyFurman, line 9: should also be m_ and also private

src/main/java/frc/robot/commands/ExtendDeployingBar.java
```diff
@@ -0,0 +1,48 @@
...
+  /**
+   * Creates a new ExtendDeployingBar
+   *
+   * @param subsystem The subsystem used by this command.
+   */
+  public ExtendDeployingBar(DeployingBar deployingbar, double setpoint) {
```
AnthonyFurman, line 15: this is minor but you should capitalize the B in deployingbar, seems to be uncapitalized in a few places

src/main/java/frc/robot/commands/ExtendDeployingBar.java
```diff
@@ -0,0 +1,48 @@
...
+    m_deployingbar = deployingbar;
+    this.setpoint = setpoint;
+    // Use addRequirements() here to declare subsystem dependencies.
+    addRequirements(deployingbar);
+  }
+  
+  // Called when the command is initially scheduled.
+  @Override
+  public void initialize() {
+    m_deployingbar.zeroEncoders();
```
AnthonyFurman, line 25: the deploying bar might not be at zero, so don't zero the encoders.

src/main/java/frc/robot/commands/ExtendDeployingBar.java
```diff
@@ -0,0 +1,48 @@
...
+    m_deployingbar.resetPID();
+    m_deployingbar.setSetpoint(setpoint);
+    m_deployingbar.setEnableStatus(true);
+  }
+
+  // Called every time the scheduler runs while the command is scheduled.
+  @Override
+  public void execute() {
```
AnthonyFurman, line 33: you can delete execute() and end()

src/main/java/frc/robot/subsystems/DeployingBar.java
```diff
@@ -0,0 +1,71 @@
...
+  public boolean atSetpoint(){
+    return m_pid.atSetpoint();
+  }
+  
+  public double getEncoderValue(){
+    return (m_motor1.getSelectedSensorPosition() + m_motor2.getSelectedSensorPosition())/2;
+  }
+
+  public void setEnableStatus(boolean enableStatus){
+    if(enableStatus)
```
AnthonyFurman, line 56: you can just do isEnabled = enableStatus. Also, you can just call this function setEnable()

src/main/java/frc/robot/subsystems/DeployingBar.java
```diff
@@ -0,0 +1,71 @@
...
+    m_motor1 = MotorFactory.createTalonFX(DeployingBarConstants.kLeftMotor, getName());
+    m_motor2 = MotorFactory.createTalonFX(DeployingBarConstants.kRightMotor, getName());
+    m_pid = new PIDController(DeployingBarConstants.kP, DeployingBarConstants.kI, DeployingBarConstants.kD);
+    m_motor1.setNeutralMode(NeutralMode.Brake);
+    m_motor2.setNeutralMode(NeutralMode.Brake);
+  }
+  @Override
+  public void periodic() {
+    if(isEnabled){
+      m_motor1.set(ControlMode.PercentOutput, m_pid.calculate(getEncoderValue(), setpoint));
```
AnthonyFurman, line 38: If there are two motors, then maybe you should have two PIDs and they should use their own encoder values. In case one side moves slightly differently than the other.

AnthonyFurman: Also, percentOutput is the default so you don't need to specify that.

src/main/java/frc/robot/subsystems/DeployingBar.java
```diff
@@ -0,0 +1,71 @@
...
+  private final TalonFX m_motor2;
+
+  private double setpoint;
+  private PIDController m_pid;
+  private boolean isEnabled;
+
+  public DeployingBar() {
+    m_motor1 = MotorFactory.createTalonFX(DeployingBarConstants.kLeftMotor, getName());
+    m_motor2 = MotorFactory.createTalonFX(DeployingBarConstants.kRightMotor, getName());
+    m_pid = new PIDController(DeployingBarConstants.kP, DeployingBarConstants.kI, DeployingBarConstants.kD);
```
AnthonyFurman, line 31: For this subsystem, I thought there would be limit switches (or hall effect) that just detect if the bar is extended. Then we just power the bar until it hits the limit switch. A PID would work but would require more tuning.

zhoutony05: To be honest, Allan is MIA every class and they are completely redesigning the subsystem so I am just doing PID for now

zhoutony05: Kaushik is suggesting using motor current to detect this, what are your thoughts about that?

AnthonyFurman: Kind of a bruh moment wrt the extending bar CAD, I guess for now we can assume a PID, because I realized depending on how they make it, it might be hard to put a limit switch. 
I'm not sure motor current would work well, current can be noisy, current can also change if there are mechanical issues, and it would mean it would mess up if it felt any pressure (ex. hitting a wall, getting hit by another robot). Maybe we can test it, but I don't think it's a good idea.

GLRoylance:
If you use two motors, there should be a separate PID controller for each motor.

Each PID controller should use the corresponding motor's encoder (not the average encoder value). Image the situation where motor1 is at the desired setpoint and motor2 is 4 inches short. Motor1 would be told to go 2 inches further and motor2 would think it is halfway there already. Instead, motor1 should be happy and motor2 should know it still has 4 inches to go.

setSetpoint does not set the PID setpoint; it is simpler if it did. The PID setpoint is continually set during periodic(). You could even get rid of private double setpoint: use the field in the PID class.

atSetpoint should test that both PIDs are atSetpoint.

GLRoylance:
m_motor1 = MotorFactory.createTalonFX(DeployingBarConstants.kLeftMotor, getName());

getName() gets the subsystem name. It wants the appropriate CAN name.

GLRoylance:
public static final int kP = 0, kI = 0, kD = 0;

PID coefficents are doubles.

GLRoylance:
// Called when the command is initially scheduled.
  @Override
  public void initialize() {
    m_deployingbar.zeroEncoders();
    m_deployingbar.resetPID();
    m_deployingbar.setSetpoint(setpoint);
    m_deployingbar.setEnableStatus(true);
  }

Do not zero the encoders.  The subsystem should know about encoders, but commands should not know about the subsystem internals.

Do not reset the PID in a command. The subsystem may do that in its setSetpoint method.

Do not enable the subystem. It should be enabled when the command is invoked.

src/main/java/frc/robot/commands/ExtendDeployingBar.java
```diff
@@ -0,0 +1,38 @@
...
+  // Called when the command is initially scheduled.
+  @Override
+  public void initialize() {
+    m_deployingBar.setSetpoint(m_setpoint);
+    m_deployingBar.setEnable(true);
+  }
+
+  @Override
+  public void end(boolean interrupted) {
+    m_deployingBar.setEnable(false);
```
PandaNinja10, line 30: why is the extendeing bar disabled wouldn't it be useful to keep the extending bar a the set point if you acidentaly bump something

src/main/java/frc/robot/constants/DeployingBarConstants.java
```diff
@@ -0,0 +1,13 @@
+package frc.robot.constants;
+
+public class DeployingBarConstants {
+  public static final int kLeftMotor = 0;
+  public static final int kRightMotor = 0;
```
PandaNinja10, line 5: second motor is unsused

src/main/java/frc/robot/constants/DeployingBarConstants.java
```diff
@@ -0,0 +1,10 @@
+package frc.robot.constants;
+
+public class DeployingBarConstants {
+  public static final int kMotor = 0;
+  public static final double kMaxRotation = 90, kMinRotation = 0;
+  public static final double kP = 1, kI = 0, kD = 0;
```
PandaNinja10, line 6: generealy we put constants on diffrent lines

zhoutony05: There are gonna be some pretty drastic code changes since they are now using limit switches, so I'd recommend looking at this branch later

AnthonyFurman:
Yeah definitely this PR is gonna change a lot. @zhoutony05 maybe mark the PR as a draft, or just close it?

zhoutony05:
Only using limit switches for now, we'll see if pid + ff is needed

PR #14, opened 2023-02-11
Roboname

GLRoylance:
## Overview

This PR uses Preferences to determine the robot identity.

## Tests Ran

SimGUI


AnthonyFurman:
@GLRoylance , there is no classbot 5, and I changed .compareTo() instead to .equals(), because it seems like other than for null comparisons they are equivalent, I assume that is fine.

AnthonyFurman (approved):

PR #15, opened 2023-02-12
Swerve pr

PandaNinja10:
## Overview

This PR will get many basic primary features of swerve drive and many corisponding tests. 

## Design Doc

## Tests Ran

- The defult rotational drive was driveing about with no visable problems
- Heading drive is currently inacessable but functional and in the code
- Circle drive - drives robot in circle with some drift do to impersice values
- Drive and Steer Volts are both functional
- Odomatry test and Heading test works but values for positional and heading PID values are still untuned
- Drive and Single Steer Feed Forward work - Steer feed forword is genrealy inacurate due to varring outside conditions so run time is planned to be extended and/or we will have higher reliance of PID
- Drive Velocity Test seams to work with no visable problems
- Shuffleboard's modules tab is currently too cluttered so you will have to commnet and uncomment module values as needed

## Additional Notes

All constants are very rough and the acessable constants acessable by defult are test swerve cosntants. Proper costants and easy constant switching will be added in another PR soon.


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -64,8 +89,17 @@ public Command getAutonomousCommand() {
    * Adds the test commands to shuffleboard so they can be run that way.
    */
   public void addTestCommands() {
-    ShuffleboardTab tab = Shuffleboard.getTab("Test");
-    tab.add("Do Nothing", new DoNothing());
+
+    GenericEntry testEntry = m_testTab.add("Test Results", false).getEntry();
+    m_testTab.add("Circle Drive", new CircleDrive(m_drive, m_drive.getRequestedDriveVelocityEntry(), m_drive.getRequestedSteerVelocityEntry()));
```
AnthonyFurman, line 94: Is there a reason that the shuffleboard entries are passsed in even though they can be gotten from the m_drive object that CircleDrive has access to?

AnthonyFurman: Same for all of the below commands.

AnthonyFurman:
Please add more javadoc comments

src/main/java/frc/robot/subsystems/Drivetrain.java
```diff
@@ -1,65 +1,525 @@
...
+  public SwerveModuleState[] m_swerveModuleStates = new SwerveModuleState[] {
+    new SwerveModuleState(),
+    new SwerveModuleState(),
+    new SwerveModuleState(),
+    new SwerveModuleState()
+  };
+
+  private final Module[] m_modules;
+
+  // TODO: Should this be public?
```
vfurman-gh, line 51: Remove obsolete comment (it is public).

src/main/java/frc/robot/subsystems/Drivetrain.java
```diff
@@ -1,65 +1,525 @@
...
 
-    m_leftMotor1 = MotorFactory.createTalonFX(DriveConstants.kLeftMotor1, Constants.kRioCAN);
-    m_leftMotor2 = MotorFactory.createTalonFX(DriveConstants.kLeftMotor2, Constants.kRioCAN);
-    m_rightMotor1 = MotorFactory.createTalonFX(DriveConstants.kRightMotor1, Constants.kRioCAN);
-    m_rightMotor2 = MotorFactory.createTalonFX(DriveConstants.kRightMotor2, Constants.kRioCAN);
+  // Displays the field with the robots estimated pose on it
+  private final Field2d m_fieldDisplay = new Field2d();
+  
+  // PID Controllers
+  private PIDController m_xController = new PIDController(0.1, 0, 0);
```
vfurman-gh, line 73: Should 0.1 be a constant?

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -90,4 +123,9 @@ public void loadCommandSchedulerShuffleboard() {
     CommandScheduler.getInstance().onCommandInterrupt(command -> Shuffleboard.addEventMarker("Command interrupted", command.getName(), EventImportance.kNormal));
     CommandScheduler.getInstance().onCommandFinish(command -> Shuffleboard.addEventMarker("Command finished", command.getName(), EventImportance.kNormal));
   }
+
+  public void shuffleboardUpdates(){
+    m_drive.updateDriveModuleFeedforwardShuffleboard();
+    m_drive.updateDriveModuleFeedforwardShuffleboard();
```
AnthonyFurman, line 129: I think this can happen in the drivetrain's periodic(), if it is necessary. I also think there is a better way to do it than to run this periodically, like something that come to mind is putting an instant command array on shuffleboard that sets the new module. That way it is just called when the thing is changed, by clicking the button, rather than having to check every 20ms if the value changed.

src/main/java/frc/robot/subsystems/Drivetrain.java
```diff
@@ -1,65 +1,525 @@
...
+    m_driveVelocity,
+    m_steerVelocity, 
+    m_steerAngle, 
+    m_drivetrainvolts, 
+    m_driveStaticFeedforward, 
+    m_driveVelocityFeedforward, 
+    m_steerStaticFeedforward,
+    m_steerVelocityFeedforward,
+    m_heading;
+  ShuffleboardTab m_swerveModulesTab,m_drivetrainTab;
```
vfurman-gh, line 88: Add space after ','

src/main/java/frc/robot/subsystems/Drivetrain.java
```diff
@@ -1,65 +1,525 @@
...
 
-    m_leftMotor1.configSupplyCurrentLimit(supplyCurrentLimit);
-    m_leftMotor2.configSupplyCurrentLimit(supplyCurrentLimit);
-    m_rightMotor1.configSupplyCurrentLimit(supplyCurrentLimit);
-    m_rightMotor2.configSupplyCurrentLimit(supplyCurrentLimit);
+  public Double[] m_driveVelFeedForwardSaver=new Double[4];
+  public Double[] m_driveStaticFeedForwardSaver=new Double[4];
+  public Double[] m_steerVelFeedForwardSaver=new Double[4];
+  public Double[] m_steerStaticFeedForwardSaver=new Double[4];
+  // modules needed to distigue in chooser
```
vfurman-gh, line 94: Spelling?

src/main/java/frc/robot/subsystems/Drivetrain.java
```diff
@@ -1,65 +1,525 @@
...
+    m_modules = new Module[]{
+        Module.create(ModuleConstants.COMP_FL, m_swerveModulesTab),
+        Module.create(ModuleConstants.COMP_FR, m_swerveModulesTab),
+        Module.create(ModuleConstants.COMP_BL, m_swerveModulesTab),
+        Module.create(ModuleConstants.COMP_BR, m_swerveModulesTab)
+      };
+    m_prevModule = m_modules[0];
+
+    m_odometry = new SwerveDriveOdometry(m_kinematics, m_pigeon.getRotation2d(), getModulePositions(), m_robotPose);
+    m_rotationController.enableContinuousInput(-Math.PI,Math.PI);
```
vfurman-gh, line 116: Space after ,

src/main/java/frc/robot/commands/SetFormationX.java
```diff
@@ -0,0 +1,29 @@
...
+        addRequirements(drive);
+    }
+    public void execute(){
+        m_drive.m_swerveModuleStates = new SwerveModuleState[] {
+            new SwerveModuleState(0.01, new Rotation2d(Units.degreesToRadians(-45))),
+            new SwerveModuleState(0.01, new Rotation2d(Units.degreesToRadians(45))),
+            new SwerveModuleState(0.01, new Rotation2d(Units.degreesToRadians(-45))),
+            new SwerveModuleState(0.01, new Rotation2d(Units.degreesToRadians(45)))
+          };
+          m_drive.setModuleStates(m_drive.m_swerveModuleStates);
```
AnthonyFurman, line 26: doesn't setModuleStates set the module states? Why do you set them above? I'm not sure that variable should be public anyway.
Ok i have looked at the code and it seems setModuleStates, in fact, does not set the module states in the drivetrain. That should probably be changed or maybe the variable should be removed and it should be stored per module.

src/main/java/frc/robot/commands/test/DriveFeedForwardCharacterzation.java
```diff
@@ -0,0 +1,87 @@
...
+
+
+
+import edu.wpi.first.wpilibj.Timer;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.subsystems.Drivetrain;
+import frc.robot.util.FeedForwardCharacterizationData;
+
+
+/** Add your docs here. */
```
AnthonyFurman, line 15: please include more javadoc comments, like this one seems like a template comment. Additionally normal comments explaining what stuff does.

src/main/java/frc/robot/subsystems/Drivetrain.java
```diff
@@ -1,65 +1,525 @@
...
    *
-   * @param forward the commanded forward movement
-   * @param turn the commanded turn rotation
+   * @param xSpeed speed of the robot in the x direction (forward)
+   * @param ySpeed speed of the robot in the y direction (sideways)
+   * @param rot angular rate of the robot
+   * @param fieldRelative whether the provided x and y speeds are relative to the field
+   */
+  public void driveRot(double xSpeed, double ySpeed, double rot, boolean fieldRelative) {
+
```
vfurman-gh, line 168: Two empty lines, remove them?

src/main/java/frc/robot/subsystems/Drivetrain.java
```diff
@@ -1,65 +1,525 @@
...
+
+    m_swerveModuleStates =
```
vfurman-gh, line 170: For simulation to work, there need to be pigeon yaw update somewhere around this code. It is probably OK leave it as is for now, if we are planning to have a separate PR from simulation branch.

src/main/java/frc/robot/commands/test/DriveFeedForwardCharacterzation.java
```diff
@@ -0,0 +1,87 @@
...
+public class DriveFeedForwardCharacterzation extends CommandBase {
+  double value = 0;
```
AnthonyFurman, line 17: include m_, also a better name than "value" would be more clear, probably some name like m_volts, i was confused initially what this was

src/main/java/frc/robot/subsystems/Drivetrain.java
```diff
@@ -1,65 +1,525 @@
...
+    for (int i = 0; i < 4; i++) {
+      m_modules[i].setDriveVoltage(0);
+      if (module == i){
+        m_modules[i].setSteerVoltage(value);
+      } else {
+        m_modules[i].setSteerVoltage(0);
+      }
+    }
+  }
+
```
vfurman-gh, line 336: Here and below: inconsistent formatting, should be only a single empty line after  function body closing bracket.

src/main/java/frc/robot/subsystems/Drivetrain.java
```diff
@@ -1,65 +1,525 @@
...
+    setUpModuleChooser();
+    setUpFeedforwardSavers();
+    
+    // inputs
+    m_swerveModulesTab.add("Module Chooser", m_moduleChooser);
+    m_driveVelocity = m_swerveModulesTab.add("Set Drive Velocity", 0).getEntry();
+    m_steerVelocity = m_swerveModulesTab.add("Set Steer Velocity", 0).getEntry();
+    m_steerAngle = m_swerveModulesTab.add("Set Steer Angle", 0).getEntry();
+    m_drivetrainvolts = m_swerveModulesTab.add("Set Volts", 0).getEntry();
+    m_driveStaticFeedforward = m_swerveModulesTab.add("Drive kS FF", m_driveStaticFeedForwardSaver[m_moduleChooser.getSelected().getModuleType().getID()]).getEntry();
```
vfurman-gh, line 390: Here and below: indent  is needed if function call is spread across multiple lines.

src/main/java/frc/robot/subsystems/Drivetrain.java
```diff
@@ -1,65 +1,525 @@
...
+    m_driveVelocityFeedforward = m_swerveModulesTab.add("Drive kV FF", m_driveVelFeedForwardSaver[m_moduleChooser.getSelected().getModuleType().getID()]).getEntry();
+    m_steerStaticFeedforward = m_swerveModulesTab.add("Steer kS FF", m_steerStaticFeedForwardSaver[m_moduleChooser.getSelected().getModuleType().getID()]).getEntry();
+    m_steerVelocityFeedforward = m_swerveModulesTab.add("Steer kV k FF", m_steerVelFeedForwardSaver[m_moduleChooser.getSelected().getModuleType().getID()]).getEntry();
+    
+
```
vfurman-gh, line 395: multiple  empty lines

src/main/java/frc/robot/commands/test/DriveVoltage.java
```diff
@@ -0,0 +1,28 @@
+package frc.robot.commands.test;
+
+import edu.wpi.first.networktables.GenericEntry;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.subsystems.Drivetrain;
+
+public class DriveVoltage extends CommandBase {
```
AnthonyFurman, line 7: I'm not sure this is necessary, it is put on the test tab but it doesn't seem to test anything, other than the fact that the motors would spin, which seems redundant with other tests.

src/main/java/frc/robot/commands/test/SteerFeedForwardCharacterzationSingle.java
```diff
@@ -0,0 +1,81 @@
...
+
+import edu.wpi.first.wpilibj.Timer;
+import edu.wpi.first.wpilibj.smartdashboard.SendableChooser;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.subsystems.Drivetrain;
+import frc.robot.subsystems.Module;
+import frc.robot.util.FeedForwardCharacterizationData;
+
+
+/** Add your docs here. */
```
vfurman-gh, line 17: Add a comment what this is doing and how? E.g, "Computes feedforward coefficients kS and kV  for Swerve steer motors. It sends  increasing voltage to the motors (in steps of 0.2 V)  and measures velocity. When voltage reaches  6V , it stops and computes linear regression of measured velocities as a function of voltage."

src/main/java/frc/robot/commands/test/SteerFeedForwardCharacterzationSingle.java
```diff
@@ -0,0 +1,81 @@
...
+public class SteerFeedForwardCharacterzationSingle extends CommandBase {
```
vfurman-gh, line 18: Rename  value -> m_voltage

src/main/java/frc/robot/commands/test/SteerFeedForwardCharacterzationSingle.java
```diff
@@ -0,0 +1,81 @@
...
+    m_timer.start();
+    m_drive.setAllOptimize(false);
+    m_feedForwardCharacterizationData = new FeedForwardCharacterizationData();
+    this.m_module = m_moduleChooser.getSelected();
+  }
+
+  public void execute() {
+    runCharacterizationVolts();
+    if (m_timer.get() > 0.5) {
+      m_feedForwardCharacterizationData.add(m_module.getSteerVelocity(), value);
```
vfurman-gh, line 43: Can we characterize all 4 motors simultaneously by keeping  a 4 element array of FeedForwardCharacterizationData and recording the data in appropriate places? Then, there is no need to have a module chooser, and all procedure will go faster.

src/main/java/frc/robot/commands/test/SteerFeedForwardCharacterzationSingle.java
```diff
@@ -0,0 +1,81 @@
...
+  public void end(boolean interrupted) {
+    System.out.println("FINISHED");
+ 
+    m_feedForwardCharacterizationData.print();
+  
+    m_drive.getSteerStaticFeedforwardArray()[m_module.getModuleType().getID()] = m_feedForwardCharacterizationData.getStatic();
+    m_drive.getSteerVelocityFeedforwardArray()[m_module.getModuleType().getID()] = m_feedForwardCharacterizationData.getVelocity();
+    System.out.println("Static " + ": " + m_feedForwardCharacterizationData.getStatic());
+    System.out.println("Velocity " + ": " + m_feedForwardCharacterizationData.getVelocity());
+
```
vfurman-gh, line 69: Please, print  m_feedForwardCharacterizationData.R() to be able to tell how well the regression matches the data.

src/main/java/frc/robot/commands/test/SteerFeedForwardCharacterzationSingle.java
```diff
@@ -0,0 +1,81 @@
...
+    m_module.setDriveVoltage(0);
+    m_module.setSteerVoltage(0);
+    
+    m_drive.getSteerStaticFeedforwardEntry().setDouble(m_drive.getSteerStaticFeedforwardArray()[m_moduleChooser.getSelected().getModuleType().getID()]);
+    m_drive.getSteerVelocityFeedforwardEntry().setDouble(m_drive.getSteerVelocityFeedforwardArray()[m_moduleChooser.getSelected().getModuleType().getID()]);
+  }
+
+  public boolean isFinished() {
+    //System.out.println(value > 11);
+    return value>6;
```
vfurman-gh, line 79: Introduce and use kMaxVoltage constant?

src/main/java/frc/robot/commands/test/DriveFeedForwardCharacterzation.java
vfurman-gh, line 17: Rename to m_voltage

src/main/java/frc/robot/commands/test/DriveFeedForwardCharacterzation.java
vfurman-gh, line 15: Add comments, same as I asked in SteerFeedforward file

src/main/java/frc/robot/commands/test/DriveFeedForwardCharacterzation.java
```diff
@@ -0,0 +1,87 @@
...
+  FeedForwardCharacterizationData[] m_feedForwardCharacterizationData;
+  
+
+  Timer m_timer = new Timer();
+  Drivetrain m_drive;
+
+  public DriveFeedForwardCharacterzation(Drivetrain drive) {
+
```
vfurman-gh, line 25: Remove emepty line

src/main/java/frc/robot/commands/test/DriveFeedForwardCharacterzation.java
```diff
@@ -0,0 +1,87 @@
...
+  public void end(boolean interrupted) {
+    System.out.println("FINISHED");
+    for (int i=0; i<4; i++){
+      m_feedForwardCharacterizationData[i].print();
+    }
+    
+    for (int i=0; i<4;i++){
+      m_drive.getDriveStaticFeedforwardArray()[i] = m_feedForwardCharacterizationData[i].getStatic();
+      m_drive.getDriveVelocityFeedforwardArray()[i] = m_feedForwardCharacterizationData[i].getVelocity();
+      System.out.println("Static " + i + ": " + m_feedForwardCharacterizationData[i].getStatic());
```
vfurman-gh, line 71: Print R() value?

src/main/java/frc/robot/RobotContainer.java
PandaNinja10: no not really in fact the single steer feedforward  gets the module chooser though m_drive. either way it is changed now. ovoius the test enrty how ever is on a diffrent tab and not stored in drive so it is staying.

src/main/java/frc/robot/subsystems/Drivetrain.java
PandaNinja10: it is a dummy P value it is too small but lets you run the test

src/main/java/frc/robot/commands/test/DriveFeedForwardCharacterzation.java
```diff
@@ -0,0 +1,87 @@
...
+      System.out.println("Velocity " + i + ": " + m_feedForwardCharacterizationData[i].getVelocity());
+    }
+    
+    m_drive.getDriveStaticFeedforwardEntry().setDouble(m_drive.getDriveStaticFeedforwardArray()[m_drive.getModuleChooser().getSelected().getModuleType().getID()]);
```
vfurman-gh, line 76: Maybe update all modules, then there is no need for module chooser?

src/main/java/frc/robot/RobotContainer.java
PandaNinja10: moved to perodic for now. was thinking we could call shuffleboard modules automaticly per test. also instant commands will coss problems if someone calls it twice on acident.

src/main/java/frc/robot/commands/test/DriveFeedForwardCharacterzation.java
```diff
@@ -0,0 +1,87 @@
...
+    m_drive.getDriveVelocityFeedforwardEntry().setDouble(m_drive.getDriveVelocityFeedforwardArray()[m_drive.getModuleChooser().getSelected().getModuleType().getID()]);
+    
+    m_drive.stop();
+  }
+
+  public boolean isFinished() {
+    //System.out.println(value > 11);
+    return value > 11;
```
vfurman-gh: Named constant?

src/main/java/frc/robot/commands/SetFormationX.java
PandaNinja10: I was very confused what the variable was used for but good to see that I am not alone. In fact I already stored it per module becuse it so much easier in terms of accessing.

src/main/java/frc/robot/subsystems/Drivetrain.java
PandaNinja10: I intentialy removed most if not all of the simulation features as to have it on a sepreate PR so they will be added later

src/main/java/frc/robot/subsystems/Drivetrain.java
PandaNinja10: In consistent sure, but there is a wider probolem of no clear standard line spaceing wich is something to discuss

src/main/java/frc/robot/subsystems/Drivetrain.java
PandaNinja10: are you saying that the commands hould be spread across multiple lines?

src/main/java/frc/robot/commands/test/DriveVoltage.java
PandaNinja10: This and Steer volts are more legeacy test so we could remove them but rn they are still there

src/main/java/frc/robot/commands/test/SteerFeedForwardCharacterzationSingle.java
PandaNinja10: Right now we decided that running them simultaneously causes too much noise to characterize each one properly and running each one right after another exists but isn't in PR as it sometimes randomly spits a out of index error

src/main/java/frc/robot/subsystems/Drivetrain.java
```diff
@@ -1,65 +1,525 @@
...
+    
+    
+    for (int i = 0; i < 4; i++){
+      m_modules[i].setupModulesShuffleboard();
+    }
+  }
+  private void setUpFeedforwardSavers(){
```
vfurman-gh, line 389: The saver functionality introduces significant complexity to this module. Do we actually need it?  Why not simply update the constants based on the characterize output?

src/main/java/frc/robot/commands/test/DriveFeedForwardCharacterzation.java
PandaNinja10: we are updating an entry not a module so we can't do it that way. On a similar note the reason why we have a chooser for feed forward values is so we don't clutter the tab with quadruple the entries.

src/main/java/frc/robot/subsystems/Drivetrain.java
PandaNinja10: the reason why we put it on shuffleboard is so we can copy paste it without using the log as it can be kind of finicky

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -23,33 +27,53 @@
...
+  private final Operator m_operator = new Operator();
+
   /** The container for the robot. Contains subsystems, OI devices, and commands. */
   public RobotContainer() {
 
     // This is really annoying so it's disabled
     DriverStation.silenceJoystickConnectionWarning(true);
 
-    // load paths before auto starts
-    PathGroupLoader.loadPathGroups();
```
AnthonyFurman, line 39: The paths should still be loaded

src/main/java/frc/robot/subsystems/Drivetrain.java
vfurman-gh: Can you make it a constant and put  comment for it saying exactly what you said above?

src/main/java/frc/robot/subsystems/Drivetrain.java
vfurman-gh: OK,  looks you strongly prefer to use Shuffleboard to output characterization results.  But then, do we have to have these savers to be inside  Drivetrain.java? I suggest we put them in the characterization code itself, or inside the callers of the charaterization command.

src/main/java/frc/robot/commands/test/DriveFeedForwardCharacterzation.java
vfurman-gh: Why can't we use SmartDashboard to output all the values we need?

AnthonyFurman: It might take more time, but a prefered solution would be some way to compactly display all of the values. Look into https://docs.wpilib.org/en/stable/docs/software/dashboards/shuffleboard/layouts-with-code/index.html
or
https://docs.wpilib.org/en/stable/docs/software/dashboards/shuffleboard/custom-widgets/index.html
However that can be a separate PR

src/main/java/frc/robot/commands/test/DriveVoltage.java
AnthonyFurman: I think then they should be removed. Same with the circle drive and other tests. If they don't have any purpose then they should be removed, and also the tests are all meant to test something, those commands do not do that right?

src/main/java/frc/robot/commands/test/OdometryTestCommand.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+  
+  @Override
+  public void execute() {
+    m_drive.runChassisPID(m_finalPose.getX(), m_finalPose.getY(), m_finalPose.getRotation().getRadians()); 
+  }
+
+  @Override
+  public boolean isFinished() {
+    // TODO: the current PID values don't allow the command to finish
```
AnthonyFurman, line 41: Should we include a command that doesn't work?

src/main/java/frc/robot/util/FeedForwardCharacterizationData.java
```diff
@@ -0,0 +1,39 @@
...
+  private final List<Double> voltageData = new LinkedList<>();
+  
+  public void add(double velocity, double voltage) {
+    if (Math.abs(velocity) > 1E-4) {
+      velocityData.add(Math.abs(velocity));
+      voltageData.add(Math.abs(voltage));
+    }
+  }
+  
+  public void print() {
```
vfurman-gh, line 19: Reanme this, since it is not actually printing anything.

src/main/java/frc/robot/commands/test/SteerVoltage.java
```diff
@@ -0,0 +1,25 @@
+package frc.robot.commands.test;
+
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.subsystems.Drivetrain;
+
+public class SteerVoltage extends CommandBase {
```
AnthonyFurman, line 6: Like above, this is not a test and I'm not sure it should be included.

src/main/java/frc/robot/commands/test/TestDriveVelocity.java
```diff
@@ -0,0 +1,38 @@
...
+    m_drive.setAllOptimize(false);
+    m_drive.setModuleStates(new SwerveModuleState[] {
+      new SwerveModuleState(m_drive.getRequestedDriveVelocityEntry().getDouble(0), new Rotation2d(Units.degreesToRadians(135))),
+      new SwerveModuleState(m_drive.getRequestedDriveVelocityEntry().getDouble(0), new Rotation2d(Units.degreesToRadians(45))),
+      new SwerveModuleState(m_drive.getRequestedDriveVelocityEntry().getDouble(0), new Rotation2d(Units.degreesToRadians(225))),
+      new SwerveModuleState(m_drive.getRequestedDriveVelocityEntry().getDouble(0), new Rotation2d(Units.degreesToRadians(315)))
+    });
+    m_testEntry.setBoolean(m_drive.isDriveVelocityAccurate());
+  }
+
```
AnthonyFurman, line 32: The point of the test is to have an isFinished() and that it reaches the velocity within a certain timeframe

src/main/java/frc/robot/commands/test/TestSteerAngle.java
```diff
@@ -0,0 +1,37 @@
...
+    m_drive.setAllOptimize(false);
+    m_drive.setModuleStates(new SwerveModuleState[] {
+      new SwerveModuleState(0.01, new Rotation2d(m_drive.getRequestedSteerVelocityEntry().getDouble(0))),
+      new SwerveModuleState(0.01, new Rotation2d(m_drive.getRequestedSteerVelocityEntry().getDouble(0))),
+      new SwerveModuleState(0.01, new Rotation2d(m_drive.getRequestedSteerVelocityEntry().getDouble(0))),
+      new SwerveModuleState(0.01, new Rotation2d(m_drive.getRequestedSteerVelocityEntry().getDouble(0)))
+    });
+    m_testEntry.setBoolean(m_drive.isSteerAngleAccurate());
+  }
+
```
AnthonyFurman, line 31: Same as for the drive test, this needs to test if the set angle is accurate.

src/main/java/frc/robot/commands/test/TestSteerAngle.java
```diff
@@ -0,0 +1,37 @@
...
+  }
+  
+  @Override
+  public void execute() {
+    m_drive.setAllOptimize(false);
+    m_drive.setModuleStates(new SwerveModuleState[] {
+      new SwerveModuleState(0.01, new Rotation2d(m_drive.getRequestedSteerVelocityEntry().getDouble(0))),
+      new SwerveModuleState(0.01, new Rotation2d(m_drive.getRequestedSteerVelocityEntry().getDouble(0))),
+      new SwerveModuleState(0.01, new Rotation2d(m_drive.getRequestedSteerVelocityEntry().getDouble(0))),
+      new SwerveModuleState(0.01, new Rotation2d(m_drive.getRequestedSteerVelocityEntry().getDouble(0)))
```
AnthonyFurman, line 27: Should the drive velocity be set here? That might make it drive into something

src/main/java/frc/robot/constants/DriveConstants.java
```diff
@@ -1,9 +1,175 @@
...
+    public static final double kSteerKVBackRight = 0.5;
+    public static final double kSteerPBackRight = 12;
+    public static final double kSteerIBackRight = 0;
+    public static final double kSteerDBackRight = 0;
+    
+  }
+  
+  // PID
+  // Drive
+  public static final double kDriveP = 0;
```
AnthonyFurman, line 151: Seems like these variables and below are already defined above in the CompBotConstants or TestDriveConstants?

src/main/java/frc/robot/commands/test/TestHeadingPID.java
```diff
@@ -0,0 +1,40 @@
...
+  // TODO: return if in error bound
+  public TestHeadingPID(Drivetrain drive){
+    m_drive = drive;
+    addRequirements(m_drive);
+  }
+  
+  @Override
+  public void execute() {
+    m_drive.setAllOptimize(false);
+    m_drive.m_headingPIDOutput = m_drive.getRotationController().calculate(m_drive.getAngleHeading(), m_drive.getRequestedHeadingEntry().getDouble(0));
```
AnthonyFurman, line 23: we might use the heading PID for other stuff, and this should be a function in the drivetrain, rather than being public

src/main/java/frc/robot/commands/test/SteerFeedForwardCharacterzationSingle.java
PandaNinja10: oops, meant to put this comment on the review above

src/main/java/frc/robot/commands/test/SteerFeedForwardCharacterzationSingle.java
PandaNinja10: Right now we decided that running them simultaneously causes too much noise to characterize each one properly and running each one right after another exists but isn't in PR as it sometimes randomly spits a out of index error

src/main/java/frc/robot/commands/test/OdometryTestCommand.java
PandaNinja10: we can just increase the error margin and it will work but that is not the parameters we have on the design doc so I don't think it would be the best idea. I think it is fine to include the test and then have tuned PID constants in the next PR.

src/main/java/frc/robot/subsystems/Drivetrain.java
vfurman-gh: I'm saying that if you split a single statement (in this case function call) across multiple lines, identation should be added to lines after the first:

src/main/java/frc/robot/commands/test/TestSteerAngle.java
PandaNinja10: if you don't put a drive velocity the modules won't move at all so you have to put a drive speed that is so small it won't move. But, now resolved with a way to disable the "deadband".

src/main/java/frc/robot/commands/test/DriveFeedForwardCharacterization.java
```diff
@@ -0,0 +1,90 @@
...
+public class DriveFeedForwardCharacterization extends CommandBase {
+
+  private double m_voltage = 0;
+  private FeedForwardCharacterizationData[] m_feedForwardCharacterizationData;
+  
+  private Timer m_timer = new Timer();
+  private Drivetrain m_drive;
+  
+  /**
+   * Creates a new command.
```
AnthonyFurman, line 23: uh, technically true, but so does every other command. You have a comment for the class which would work for the constructor as you are making that class so for most use cases it would be what it does.

src/main/java/frc/robot/constants/DriveConstants.java
```diff
@@ -1,9 +1,156 @@
...
+  
+  public static final double kWheelRadius = Units.inchesToMeters(2);
+  
+  public static final double kTrackWidth = Units.inchesToMeters(20.75);//22.75 swerve bot, 20.75 comp bot
+  public static final double kDriveGearRatio = 6.75;
+  public static final double kSteerGearRatio = 12.8;
+  
+  public static final double kMaxSpeed = FalconConstants.kMaxRpm / 60.0 / kDriveGearRatio * kWheelRadius * 2 * Math.PI;
+  
+  public static final double kMaxAngularSpeed = FalconConstants.kMaxRpm / 60.0 / kSteerGearRatio * 2 * Math.PI; // 8.3 rot/s
```
AnthonyFurman, line 18: This is not the max angular speed of the robot. For that, just take the max speed of the robot and divide by the radius (v = w/r).

src/main/java/frc/robot/subsystems/Drivetrain.java
PandaNinja10: I think this is resolved but not sure as the line it is on has probably moved

src/main/java/frc/robot/subsystems/Drivetrain.java
PandaNinja10: Because the modules need the entry constantly we can't put the shuffleboard entry or savers in the command or the caller.

src/main/java/frc/robot/util/TimeAccuracyTest.java
```diff
@@ -0,0 +1,62 @@
...
+  public boolean calculate(){
+    if (m_setpointUpdateTime + m_timeMargin <= WPIUtilJNI.now() * 1e-6) m_lastUseableResult = m_accuracyTest.getAsBoolean();
+    return m_lastUseableResult;
+  }
+
+  /**
+   * Converts the double suppliers to a boolean.
+   * @return true if the actual value is within the error margin of the setpoint, false if not
+   */
+  private boolean getDoubleAccuracyTest(){
```
AnthonyFurman, line 59: probably should pass in actual and setpoint. No reason to store them as fields then

src/main/java/frc/robot/util/PolynomialRegression.java
```diff
@@ -0,0 +1,206 @@
...
+    }
+    return 0;
+  }
+
+  /**
+   * Unit tests the {@code PolynomialRegression} data type.
+   *
+   * @param args the command-line arguments
+   */
+  public static void main(String[] args) {
```
AnthonyFurman, line 199: This should be moved into a unit test instead

build.gradle
```diff
@@ -1,6 +1,6 @@
 plugins {
     id "java"
-    id "edu.wpi.first.GradleRIO" version "2023.1.1"
+    id "edu.wpi.first.GradleRIO" version "2023.2.1"
```
richtan, line 3: Newest version is 2023.3.2

AnthonyFurman (approved):

PR #16, opened 2023-02-12
Angled elevator

lakshag32:
This PR will add code to control the angled elevator on this year's robot. 

Design Doc: https://docs.google.com/document/d/1BCxfgVRcpiqJ0dqkICla7OOI6J3N9_fGweoOEehTh8E/edit?usp=sharing

Tests Ran:
None for now.


src/main/java/frc/robot/Robot.java
```diff
@@ -18,6 +18,7 @@
  * project.
  */
 public class Robot extends TimedRobot {
+  //TODO: Figure out how to execute commands
```
AnthonyFurman, line 21: Sorry, I'm not sure what this means?

src/main/java/frc/robot/commands/ManualElevatorControl.java
```diff
@@ -0,0 +1,34 @@
...
+
+  @Override
+  public void end(boolean interrupted) {
+    m_elevator.stopMotor();
+  }
+
+  // Returns true when the command should end.
+  @Override
+  public boolean isFinished() {
+    if(m_elevator.returnBottomLimSwitchCondition() == true || m_elevator.returnTopLimSwitchCondition() == true){
```
AnthonyFurman, line 29: If this is meant to be a manual control command, it probably should never finish. Otherwise they will not be able to control the elevator.

src/main/java/frc/robot/commands/MoveElevatorRecordMaxHeight.java
```diff
@@ -0,0 +1,33 @@
...
+
+  @Override
+  public void initialize() {
+    m_elevator.set(ElevatorConstants.kElevatorMotorEncoderZeroingPower);
+  }
+
+  @Override
+  public void end(boolean interrupted) {
+    m_elevator.stopMotor();
+    ElevatorConstants.kElevatorTopHeightMeters = m_elevator.getElevatorHeightMeters();
```
AnthonyFurman, line 22: If you are changing the value, it is probably not a constant, maybe it should just be a field in Elevator.java

src/main/java/frc/robot/commands/ElevatorCalibrationAbsoluteEncoder.java
```diff
@@ -0,0 +1,34 @@
...
+  public void end(boolean interrupted) {
+    m_elevator.stopMotor();
+    m_elevator.calibrationZeroAbsEncoder();;
+  }
+  
+  // Returns true when the command should end.
+  @Override
+  public boolean isFinished() {
+    if(m_elevator.returnBottomLimSwitchCondition() == true){
+      return true;
```
AnthonyFurman, line 29: for these kinds of if statements, you can just do `return m_elevator.returnBottomLimSwitchCondition()` because it will be the same as `return true` or `return false`

AnthonyFurman: same for all of the other commands

src/main/java/frc/robot/commands/MoveToHeight.java
```diff
@@ -0,0 +1,23 @@
...
+
+  @Override
+  public void execute() {
+    System.out.println(m_elevator.returnHeightError(m_elevatorHeight)); 
+    m_elevator.set(m_elevator.returnClampedElevatorPID(m_elevatorHeight)); 
+    m_elevator.stopMotorsIfLimitSwitchesTripped(m_elevator.returnClampedElevatorPID(m_elevatorHeight));
+  }
+}
+
+//Note: this command will never stop running, the motor needs to constantly output power to maintain the elevator's height
```
AnthonyFurman, line 23: Notes like this should be above some relevant code. Also, the way we typically write PID is to have it in the periodic method of the subsystem. So, you can just set the elevator setpoint and the PID will be running in the subsystem.

src/main/java/frc/robot/constants/Constants.java
```diff
@@ -38,4 +38,5 @@ public final class Constants {
    */
   @Deprecated
   public static final FalconConstants falcon = new FalconConstants();
+  public static final ElevatorConstants elevator = new ElevatorConstants();
```
AnthonyFurman, line 41: You don't need to add this

src/main/java/frc/robot/constants/ElevatorConstants.java
```diff
@@ -0,0 +1,30 @@
...
+  public static final double kElevatorHeightBottomNodeMeters = -1; //based on Arnav's calcs: 
+  public static double kElevatorTopHeightMeters = 0; 
+  public static final double kElevatorBottomHeightMeters = 0; 
+  public static final double kElevatorDistPerMotorEncoderTickMeters = (((2.0 * Math.PI * Units.inchesToMeters(0.6875))/12.8571429) / 2048); //TODO: Check if this calculation is correct
+  public static final double kSpoolAbsEncoderDistancePerTickMeters = ((2.0*Math.PI* Units.inchesToMeters(0.6875))/2048);
+  public static double kElevatorAbsEncoderZeroPositionMeters = 0;
+  public static final double kMiddleNodeSetpoint = -1; 
+  public static final double kTopNodeSetpoint = -1; 
+  public static final double kElevatorMotorEncoderZeroingPower = -0.25; //TODO: This may need to be changed. 
+  public static final double kElevatorMotorMaxHeightRecordingPower = 0.25; //TODO: This may need to be changed.
```
AnthonyFurman, line 22: A lot of these constants are called kElevatorXYZ, but it is in ElevatorConstants.java so that is not necessary. You can just do ElevatorConstants.kP or kMotor, etc. Also, I think there are two motors for the elevator.

src/main/java/frc/robot/constants/ElevatorConstants.java
```diff
@@ -0,0 +1,30 @@
...
+  public static final int kElevatorMotor = -1; 
+  public static final double kElevatorP = -1; 
+  public static final double kElevatorI = -1; 
+  public static final double kElevatorD = -1; 
+  public static final double kELevatorHeightMiddleNodeMeters = -1; //based on Arnav's calcs: 
+  public static final double kELevatorHeightTopNodeMeters = -1; //based on Arnav's calcs: 
+  public static final double kElevatorHeightBottomNodeMeters = -1; //based on Arnav's calcs: 
+  public static double kElevatorTopHeightMeters = 0; 
+  public static final double kElevatorBottomHeightMeters = 0; 
+  public static final double kElevatorDistPerMotorEncoderTickMeters = (((2.0 * Math.PI * Units.inchesToMeters(0.6875))/12.8571429) / 2048); //TODO: Check if this calculation is correct
```
AnthonyFurman, line 16: Maybe to make this calculation more clear, replace 2048 with FalconConstants.ticksPerRotation, 12.8571429 with kGearRatio, and 0.6875 with kSpoolRadius (I assume that is what that is)

src/main/java/frc/robot/subsystems/Elevator.java
```diff
@@ -0,0 +1,107 @@
...
+import edu.wpi.first.wpilibj.DutyCycleEncoder;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import frc.robot.constants.Constants;
+import frc.robot.constants.ElevatorConstants;
+import frc.robot.controls.Operator;
+import frc.robot.util.MotorFactory;
+
+public class Elevator extends SubsystemBase {
+
+  private final WPI_TalonFX m_elevatorMotor; //do I define these are private final?
```
AnthonyFurman, line 23: yes it is fine to define them as private final

src/main/java/frc/robot/subsystems/Elevator.java
```diff
@@ -0,0 +1,107 @@
...
+  private final DigitalInput m_topLimitSwitch; 
+  private final PIDController m_elevatorPID; 
+  private final DutyCycleEncoder m_absoluteSpoolEncoder; 
+
+  public Elevator() {
+    this(
+      MotorFactory.createTalonFX(ElevatorConstants.kElevatorMotor, Constants.kRioCAN),
+      new DigitalInput(-1),
+      new DigitalInput(-1),
+      new DutyCycleEncoder(-1)
```
AnthonyFurman, line 34: Last year we made the subsystems this way because we had to mock them. I'm not sure that's necessary this year. Also, the limit switch ports and encoder ports should be constants.

src/main/java/frc/robot/subsystems/Elevator.java
```diff
@@ -0,0 +1,107 @@
...
+    );
+  }
+
+  public Elevator(WPI_TalonFX motor, DigitalInput bottomSwitch, DigitalInput topSwitch, DutyCycleEncoder absoluteSpoolEncoder) {
+    m_elevatorMotor = motor;
```
AnthonyFurman, line 39: could just be called m_motor

src/main/java/frc/robot/subsystems/Elevator.java
```diff
@@ -0,0 +1,107 @@
...
+
+  public void set(double power){
+    m_elevatorMotor.set(power); 
+  }
+
+  public void stopMotor(){
+    m_elevatorMotor.set(0); 
+  }
+
+  public boolean returnBottomLimSwitchCondition(){
```
AnthonyFurman, line 55: I would use "get" instead of "return", and I'm not sure "condition" is necessary.

AnthonyFurman: same for returnTopLimSwitchCondition()

src/main/java/frc/robot/subsystems/Elevator.java
```diff
@@ -0,0 +1,107 @@
...
+
+  public boolean returnTopLimSwitchCondition(){
+    return m_topLimitSwitch.get(); 
+  }
+
+  public void resetMotorEncoder(){
+    m_elevatorMotor.setSelectedSensorPosition(0); 
+  }
+
+  public void stopMotorsIfLimitSwitchesTripped(double power){
```
AnthonyFurman, line 67: The two if cases are unnecessary. Also, I think this should just be the default behavior of the set() function. I would talk to joshua about the wiring of the limit switches (or hall effect sensors), and if they can be inputted into the falcon. That would allow you to set it up so the motor automatically does not power forward if the forward limit switch is enabled. See here for documentation: https://v5.docs.ctr-electronics.com/en/latest/ch13_MC.html#limit-switches

src/main/java/frc/robot/subsystems/Elevator.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+    if (m_bottomLimitSwitch.get() && power>0){
+      //the semicolon makes the if statement do nothing
+      ;//TODO: Maybe remove this if unncesary     
+    }
+  }
+ 
+
+  public double returnHeightError(double elevatorHeightDesired){
+    double error = elevatorHeightDesired-getElevatorHeightMeters();
```
AnthonyFurman, line 86: if getElevatorHeightMeters is public, then I'm not sure this is necessary. any class can do subtraction

src/main/java/frc/robot/subsystems/Elevator.java
```diff
@@ -0,0 +1,107 @@
...
+ 
+
+  public double returnHeightError(double elevatorHeightDesired){
+    double error = elevatorHeightDesired-getElevatorHeightMeters(); 
+    return error; 
+  }
+
+  public double returnClampedElevatorPID(double elevatorHeightDesired){
+    double pidValue = MathUtil.clamp(m_elevatorPID.calculate(getElevatorHeightMeters(), elevatorHeightDesired ),-0.25,0.25);
+       //TODO: Increase clamping range to make motor go faster if needed
```
AnthonyFurman, line 92: I would have the clamp range as a constant

AnthonyFurman: Also, I would move this code to run in periodic() and have a setPosition() function that sets the PID setpoint

src/main/java/frc/robot/subsystems/Elevator.java
```diff
@@ -0,0 +1,107 @@
...
+
+    return pidValue; 
+  }
+
+  public double getElevatorHeightMeters() {
+    double elevatorHeight = m_elevatorMotor.getSelectedSensorPosition()*ElevatorConstants.kElevatorDistPerMotorEncoderTickMeters;
+    return elevatorHeight;
```
AnthonyFurman, line 99: This can be one line

src/main/java/frc/robot/subsystems/Elevator.java
```diff
@@ -0,0 +1,107 @@
...
+    );
+  }
+
+  public Elevator(WPI_TalonFX motor, DigitalInput bottomSwitch, DigitalInput topSwitch, DutyCycleEncoder absoluteSpoolEncoder) {
+    m_elevatorMotor = motor; 
+    m_bottomLimitSwitch = bottomSwitch; 
+    m_topLimitSwitch = topSwitch;
+    m_absoluteSpoolEncoder = absoluteSpoolEncoder; 
+
+    m_elevatorPID = new PIDController(ElevatorConstants.kElevatorP, ElevatorConstants.kElevatorI, ElevatorConstants.kElevatorD);
```
AnthonyFurman, line 44: The absolute encoder will keep it's position when the robot is turned off and on, but the motor encoder is not. So here in the initialization would be a good place to set the motor encoder to the same value as the absolute encoder.

src/main/java/frc/robot/commands/ResetEncoderAtBottom.java
```diff
@@ -0,0 +1,33 @@
+package frc.robot.commands;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.constants.ElevatorConstants;
+import frc.robot.subsystems.Elevator;
+
+public class ResetEncoderAtBottom extends CommandBase {
```
AnthonyFurman, line 6: I think it would make sense to combine this and the ElevatorCalibrateAbsoluteEncoder into just CalibrateElevatorZero. Is there a scenario where you would want to calibrate the motor encoder but not the absolute encoder? The absolute encoder is mainly just to store the position of the elevator even when the robot is off.

src/main/java/frc/robot/subsystems/Elevator.java
```diff
@@ -0,0 +1,107 @@
...
+import frc.robot.constants.ElevatorConstants;
+import frc.robot.controls.Operator;
+import frc.robot.util.MotorFactory;
+
+public class Elevator extends SubsystemBase {
+
+  private final WPI_TalonFX m_elevatorMotor; //do I define these are private final? 
+  private final DigitalInput m_bottomLimitSwitch;
+  private final DigitalInput m_topLimitSwitch; 
+  private final PIDController m_elevatorPID;
```
AnthonyFurman, line 26: I know feedforward was planned for this, I assume it will be in a later commit? If so, I would prioritize that.

src/main/java/frc/robot/commands/ManualElevatorControl.java
lakshag32: Yeah you are right. I will fix that.

src/main/java/frc/robot/Robot.java
lakshag32: Hi Anthony, I was just trying a todo for me to figure out how to run commands when the operator presses buttons.

src/main/java/frc/robot/commands/ManualElevatorControl.java
lakshag32: Fixed

src/main/java/frc/robot/commands/MoveElevatorRecordMaxHeight.java
lakshag32: Change has been incorporated.

src/main/java/frc/robot/commands/ElevatorCalibrationAbsoluteEncoder.java
lakshag32: Change Incorporated

src/main/java/frc/robot/constants/Constants.java
lakshag32: Change Incorporated

src/main/java/frc/robot/constants/ElevatorConstants.java
lakshag32: Change Incorporated

src/main/java/frc/robot/constants/ElevatorConstants.java
lakshag32: Done

src/main/java/frc/robot/subsystems/Elevator.java
lakshag32: Okay cool

src/main/java/frc/robot/subsystems/Elevator.java
lakshag32: Code Change incorporated

src/main/java/frc/robot/subsystems/Elevator.java
lakshag32: Changed added

src/main/java/frc/robot/subsystems/Elevator.java
GLRoylance: Absolutely not necessary.

Just have one constructor. It should set the member values using defined port constants.

    m_elevatorMotor = MotorFactory.createTalonFX(ElevatorConstants.kElevatorMotor, Constants.kRioCAN); 
    m_bottomLimitSwitch = new DigitalInput(kBottomLimitSwitchPort); 
    m_topLimitSwitch = new DigitalInput(kTopLimitSwitchPort);
    m_absoluteSpoolEncoder = new DutyCycleEncoder(kSpoolEncoder); 

Those assignment could happen where the member variables are declared.

The DutyCycleEncoder allows you to set an offset and a scale.
See DutyCycleEncoder.setPositionOffset() and .setDistancePerRotation().
They should be set in the constructor.

You should add a close method to close the digital ports. It can be called during testing.

    public void close() {
        // close the ports
        m_bottomLimitSwitch.close();
        m_topLimitSwitch.close();
        m_absoluteSpoolEncoder.close();
    }

src/main/java/frc/robot/subsystems/Elevator.java
GLRoylance: Why are two encoders being used? Isn't one or the other sufficient?

src/main/java/frc/robot/commands/ResetEncoderAtBottom.java
GLRoylance: I think knowledge of the limit switches should be in the subsystem. Say a zero method that gets invoked by a command. The subsystem should continually monitor the limit switches and do something appropriate when the a limit switch is tripped.

The command sets the motor power at the start, but what happens if the subsystem is running PID? (The tradition elevator subsystem runs PID all the time.) During the next call to periodic(), the subsystem would set a different power.

You have put the PID loop into the MoveToHeight() command. That confuses subsystems and commands. Say you are a subsystem. If I give you a command to move left 3 feet. You will do the moving. I should not be instructing something like a motor to push you 3 feet to the left.

Furthermore, think about how commands are used. There might be the sequential commands

MoveToHeight(0.85);
DropTheCone().

We need the MoveToHeight() to terminate when it gets to 0.85. But MoveToHeght never terminates, so we never drop the cone.

If MoveToHeight() terminated at 0.85, then it would stop and the cone would drop. But it would also stop running the PID loop, so the elevator will move from 0.85. Consider the sequence

MoveToHeight(0.85)
DoSomethingImportantForTheNext5Seconds()
DropTheCone()

We may drop the cone in the wrong place because the elevator moved.

src/main/java/frc/robot/subsystems/Elevator.java
GLRoylance: encoder scaling is wonderful:

  return m_encoder.getDistance();

src/main/java/frc/robot/subsystems/Elevator.java
lakshag32: Change added.

src/main/java/frc/robot/subsystems/Elevator.java
lakshag32: Code change added.

src/main/java/frc/robot/Robot.java
AnthonyFurman: Then please remove the comment, it's in the wrong class first off and you made the button presses, it's not to do if it's been done. It would be confusing to have random TODOs that have already been done.

lakshag32: Change added

src/main/java/frc/robot/subsystems/Elevator.java
lakshag32: Change added

src/main/java/frc/robot/commands/MoveToHeight.java
lakshag32: code change added

src/main/java/frc/robot/subsystems/Elevator.java
lakshag32: Code Change added.

src/main/java/frc/robot/subsystems/Elevator.java
lakshag32: Code change added.

src/main/java/frc/robot/subsystems/Elevator.java
lakshag32: code change added.

src/main/java/frc/robot/subsystems/Elevator.java
lakshag32: Yes, one should be sufficient. The absolute encoder is just another thing that we could use, so I added some code for it.

src/main/java/frc/robot/commands/ResetEncoderAtBottom.java
lakshag32: Sounds good, I will keep knowledge of limit switches within the subsystem.

src/main/java/frc/robot/subsystems/Elevator.java
lakshag32: Done

src/main/java/frc/robot/controls/Operator.java
```diff
@@ -1,15 +1,48 @@
...
 import frc.robot.subsystems.FourBarArm;
 import lib.controllers.GameController;
 import lib.controllers.GameController.Button;
+import lib.controllers.GameController.DPad;
+import lib.controllers.GameController.Axis;;
 
 public class Operator {
   private static GameController operator = new GameController(OIConstants.kOperatorJoy);
 
+  public static void configureControls(Elevator elevator) {
```
AnthonyFurman, line 19: Combine this function with the other configureControls() function

src/main/java/frc/robot/commands/ElevatorCalibrationAbsoluteEncoder.java
```diff
@@ -0,0 +1,31 @@
+// package frc.robot.commands;
+// import edu.wpi.first.wpilibj2.command.CommandBase;
+// import frc.robot.constants.ElevatorConstants;
+// import frc.robot.subsystems.Elevator;
+
+// public class ElevatorCalibrationAbsoluteEncoder extends CommandBase {
```
AnthonyFurman, line 6: If you're not using the class, feel free to delete it

src/main/java/frc/robot/commands/ManualElevatorControl.java
```diff
@@ -0,0 +1,36 @@
...
+    m_elevator.disablePID(); 
+  }
+  @Override
+  public void execute() {
+    m_elevator.set(Operator.getRawThrottleValue()); // rename throttlevalue function to somehting else
+    m_elevator.stopMotorsIfLimitSwitchesTripped(Operator.getRawThrottleValue());
+  }
+
+  @Override
+  public void end(boolean interrupted) {
```
AnthonyFurman, line 28: If it ends, it should re enable the PID.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -41,6 +43,7 @@ public RobotContainer() {
     PathGroupLoader.loadPathGroups();
 
     Driver.configureControls(m_drive);
+    Operator.configureControls(m_elevator);
     Operator.configureControls(m_arm);
```
AnthonyFurman, line 47: combine these functions

src/main/java/frc/robot/commands/ManualElevatorControl.java
```diff
@@ -0,0 +1,36 @@
...
+    addRequirements(m_elevator);
+  }
+  @Override
+  public void initialize() {
+    m_elevator.disablePID(); 
+  }
+  @Override
+  public void execute() {
+    m_elevator.set(Operator.getRawThrottleValue()); // rename throttlevalue function to somehting else
+    m_elevator.stopMotorsIfLimitSwitchesTripped(Operator.getRawThrottleValue());
```
AnthonyFurman, line 24: not sure what the second line means, the limit switches should already prevent the motor from spinning. Either that is done through the forwardLimit config in the TalonFx object or through an if statement in elevator.set()

src/main/java/frc/robot/commands/MoveToHeight.java
```diff
@@ -0,0 +1,36 @@
...
+    m_elevator.setMotorLimit();
+
+    m_elevator.setSepointMeters(m_elevatorSetpoint);
+
+  }
+
+  @Override
+  public void execute() {
+    System.out.println(m_elevatorSetpoint-m_elevator.getElevatorHeightMeters());
+    m_elevator.setSepointMeters(m_elevatorSetpoint);
```
AnthonyFurman, line 28: The setpoint can't change, so don't set it again. You don't need to do anything in periodic(). If you would like, you could log the error. LogManager.addDouble("Elevator/error", () -> m_pid.getSetpoint - getElevatorHeight()); would work in Elevator.java in the constructor to log the error.

src/main/java/frc/robot/commands/MoveToHeight.java
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  @Override
+  public void execute() {
+    System.out.println(m_elevatorSetpoint-m_elevator.getElevatorHeightMeters());
+    m_elevator.setSepointMeters(m_elevatorSetpoint); 
+
+  }
+
+  public void end(){
```
AnthonyFurman, line 32: You should have an isFinished to signify the command has ended, even if it does nothing on end.

src/main/java/frc/robot/commands/ManualElevatorControl.java
```diff
@@ -0,0 +1,36 @@
...
+package frc.robot.commands;
+
+import frc.robot.controls.Operator;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.subsystems.Elevator;
+
+public class ManualElevatorControl extends CommandBase {
+  Elevator m_elevator; 
+
+  public ManualElevatorControl(Elevator elevator) {
```
AnthonyFurman, line 13: For this command and others, add a javadoc comment above the constructor explaining simply what the command does.

src/main/java/frc/robot/subsystems/Elevator.java
AnthonyFurman: Sorry, are you sure you pushed? The function is still there.

src/main/java/frc/robot/subsystems/Elevator.java
```diff
@@ -0,0 +1,156 @@
...
+    
+    m_absoluteSpoolEncoder = new DutyCycleEncoder(ElevatorConstants.kAbsEncoderPort); 
+    
+    m_elevatorMotorEncoder = new TalonEncoder(m_motor); 
+    m_elevatorMotorEncoder.setDistancePerPulse(ElevatorConstants.kDistPerMotorEncoderTick);
+
+    m_elevatorPID = new PIDController(ElevatorConstants.kP, ElevatorConstants.kI, ElevatorConstants.kD);  
+  }
+
+  public void setMotorLimit(double powerLevel){
```
AnthonyFurman, line 50: Is there a reason you would be changing the clamp amount during operation? If not, you don't need to make a function to change it.

src/main/java/frc/robot/subsystems/Elevator.java
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  @Override
+  public void periodic() {
+    /**
+     * If we hit the bottom limit switch, then we must be at the zero position.
+     * Thus set the setpoint to 0, then atSetpoint() will be triggered, causing the 
+     * ResetEncoderAtBottom() command to end
+     */
+    if(getBottomLimitSwitch()){
```
AnthonyFurman, line 66: This would prevent the motor from ever powering upward once it hits the bottom limit switch. I think we talked about this in class. You can get the motor power with motor.get() and if it is positive, then it does not have to be set to zero.
But, I would prefer you use the built in TalonFX limit switch trigger to stop the motor.

src/main/java/frc/robot/subsystems/Elevator.java
```diff
@@ -0,0 +1,156 @@
...
+  private double clampLow = -ElevatorConstants.kMotorLimit; 
+  private double clampHigh = ElevatorConstants.kMotorLimit;
+  private boolean m_enabled; 
+
+  public Elevator() {
+    m_motor = MotorFactory.createTalonFX(ElevatorConstants.kMotorPort, Constants.kRioCAN);
+    //m_motor.configForwardLimitSwitchSource(LimitSwitchSource.FeedbackConnector, LimitSwitchNormal.NormallyOpen, 0);
+    //m_motor.configReverseLimitSwitchSource(LimitSwitchSource.FeedbackConnector, LimitSwitchNormal.NormallyOpen, 0);
+    m_bottomLimitSwitch = new DigitalInput(ElevatorConstants.kBottomLimitSwitchPort); 
+    m_topLimitSwitch = new DigitalInput(ElevatorConstants.kTopLimitSwitchPort);
```
AnthonyFurman, line 40: The limit switches are plugged into the talonFX, so I think the declaration would be more like the commented out lines.

AnthonyFurman: You would also change getting the limit switch to be talon.getSensorCollection().isFwdLimitSwitchClosed();
See more here: https://v5.docs.ctr-electronics.com/en/stable/ch13_MC.html#limit-switches
And then you can remove all the checks for if a limit switch is triggered to not spin the motors.

src/main/java/frc/robot/subsystems/Elevator.java
```diff
@@ -0,0 +1,156 @@
...
+    
+    m_absoluteSpoolEncoder = new DutyCycleEncoder(ElevatorConstants.kAbsEncoderPort); 
+    
+    m_elevatorMotorEncoder = new TalonEncoder(m_motor); 
+    m_elevatorMotorEncoder.setDistancePerPulse(ElevatorConstants.kDistPerMotorEncoderTick);
```
AnthonyFurman, line 45: Be careful with how you use both encoders and understand why each is necessary. The absolute encoder always tells you what angle it is facing, even if the robot is reset. But just the 0-360 angle. So if several rotations occur, it can loop on itself so it may not be enough to tell you exactly where the elevator is. The quadrature encoder in the talon does not loop, but it also does not persist on reset. So you can use the absolute encoder to persist the position, or maybe if there are limit switches that isn't necessary.

src/main/java/frc/robot/subsystems/Elevator.java
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  public void set(double power){
+    m_motor.set(power); 
+  }
+  public void enablePID(){
+    m_enabled = true; 
+  }
+  public void disablePID(){
+    m_enabled = false;
```
AnthonyFurman, line 95: these two can be one function

src/main/java/frc/robot/subsystems/Elevator.java
```diff
@@ -0,0 +1,156 @@
...
+    m_motor.set(power); 
+  }
+  public void enablePID(){
+    m_enabled = true; 
+  }
+  public void disablePID(){
+    m_enabled = false; 
+  }
+
+  public void setSepointMeters(double setPointMeters){
```
AnthonyFurman, line 98: minor typo

src/main/java/frc/robot/subsystems/Elevator.java
lakshag32: I kept it there so that we can try both using the encoders plugged into the Rio and the Falcon

src/main/java/frc/robot/subsystems/Elevator.java
lakshag32: code change added.

src/main/java/frc/robot/subsystems/Elevator.java
lakshag32: code change added.

src/main/java/frc/robot/subsystems/Elevator.java
lakshag32: Okay. Can I keep both of them for now?

src/main/java/frc/robot/subsystems/Elevator.java
lakshag32: Hi Anthony, I would like the elevator to go slower during calibration and faster during normal operation

src/main/java/frc/robot/commands/ElevatorCalibrationAbsoluteEncoder.java
lakshag32: Hi Anthony, I might use it a little later, so I'll keep it for now

src/main/java/frc/robot/controls/Operator.java
lakshag32: Code change added.

AnthonyFurman: It's still there...
maybe you didn't push?
You can check the code when you reply to my comment

src/main/java/frc/robot/subsystems/Elevator.java
AnthonyFurman: I think you can run it without the absolute encoder and reset the motor encoder with the limit switches,

src/main/java/frc/robot/subsystems/Elevator.java
AnthonyFurman: Well, for calibration, you choose the motor power it runs at, so there is no need for a clamp.
A clamp limits the min/max output (presumably of the PID). But for calibration you are setting the motor power, so there's no min/max.

src/main/java/frc/robot/subsystems/Elevator.java
AnthonyFurman: Well, to do that, you would need to change a different part of the code, so I'm not sure this function is necessary no matter where you plug the limit switches into.

AnthonyFurman:
@lakshag32 , it seems a lot of places you say "code changes added" but the code is the same. Make sure you commit and push your changes.

AnthonyFurman:
Sorry, accidentally hit the close PR button. the PR is still open

src/main/java/frc/robot/commands/ManualElevatorControl.java
lakshag32: it should not end, I'll delete end() and finished() rn

src/main/java/frc/robot/RobotContainer.java
lakshag32: code change added.

src/main/java/frc/robot/commands/MoveToHeight.java
lakshag32: code change added

src/main/java/frc/robot/subsystems/Elevator.java
lakshag32: Hi Anthony, I deleted functionality for using limit switches plugged into the RoboRio

src/main/java/frc/robot/subsystems/Elevator.java
lakshag32: I have used the built in TalonFX limit switch trigger to stop the motor. I may or may not have done it right.

src/main/java/frc/robot/subsystems/Elevator.java
lakshag32: oh right. I have changed the code so that the elevator does down with a set power for calibration

src/main/java/frc/robot/commands/ManualElevatorControl.java
lakshag32: yes, you are right. I have deleted line 24.

src/main/java/frc/robot/controls/Operator.java
lakshag32: sorry, I didn't push. Would you like me to push after I reply to each feedback?

src/main/java/frc/robot/subsystems/Elevator.java
lakshag32: I have uncommented those lines and have gotten the status of the limit switches using that function. See the functions getBottomLimitSwitch() and getTopLimitSwitch()

src/main/java/frc/robot/subsystems/Elevator.java
lakshag32: sounds good.

src/main/java/frc/robot/commands/MoveToHeight.java
lakshag32: I will figure this out a little later.

src/main/java/frc/robot/commands/ManualElevatorControl.java
lakshag32: Javadoc comments have been added.

src/main/java/frc/robot/commands/elevator/ManualElevatorControl.java
```diff
@@ -0,0 +1,56 @@
...
+      m_currentElevatorPos = m_elevator.getHeight(); 
+      m_elevator.setSetpoint(m_currentElevatorPos);
+    }
+
+    
+   
+  }
+
+  @Override
+  public void end(boolean interrupted) {
```
AnthonyFurman, line 47: You might want to reenable the PID and set the setpoint to the current position.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -40,7 +43,7 @@
  * subsystems, commands, and trigger mappings) should be declared here.
  */
 public class RobotContainer {
-
+  //TODO: Did I define all the variables correctly?
```
AnthonyFurman, line 46: not a big deal but TODOs like this are not stuff to do, I would either leave a comment on the PR itself, not in the code, or not write anything because we're gonna check that variables are right anyway.

src/main/java/frc/robot/commands/elevator/ElevatorCalibrationAbsoluteEncoder.java
```diff
@@ -0,0 +1,32 @@
+package frc.robot.commands.elevator;
```
AnthonyFurman, line 1: delete file

src/main/java/frc/robot/commands/elevator/MoveToHeight.java
```diff
@@ -0,0 +1,34 @@
...
+  public void initialize(){
+    //set the elevator PID speed to be clamped within the correct range to get it to move faster
+    //This should be done already in ResetEncoderAtBottom() but just in case something happens it's here too. 
+    m_elevator.setSetpoint(m_targetHeight); 
+    m_elevator.setEnabled(true);
+  }
+  
+  // Returns true when the command should end.
+  @Override
+  public boolean isFinished() {
```
AnthonyFurman, line 31: should finish when it is at the setpoint, though it will keep the PID active

src/main/java/frc/robot/commands/elevator/ResetEncoderAtBottom.java
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  @Override
+  public void initialize() {
+    m_elevator.setEnabled(false); 
+  }
+
+  @Override
+  public void execute(){
+    m_elevator.set(-ElevatorConstants.kCalibrationPower);
```
AnthonyFurman, line 25: you can just set it once, in initialize()

src/main/java/frc/robot/controls/ManualController.java
```diff
@@ -28,5 +34,20 @@ public void configureControls() {
     m_manual.get(Button.Y).onTrue(new InstantCommand(() -> m_arm.setMotorPower(0.05)));
     m_manual.get(Button.X).onTrue(new InstantCommand(() -> m_arm.setMotorPower(-0.05)));
     m_manual.get(Button.B).onTrue(new InstantCommand(() -> m_arm.setMotorPower(0)));
+
+    //Move to max height
+    m_manual.get(m_manual.LEFT_TRIGGER_BUTTON).onTrue(new MoveToHeight(m_elevator, ElevatorConstants.kMaxHeight)); 
+    //Move to min height
+    m_manual.get(m_manual.RIGHT_TRIGGER_BUTTON).onTrue(new MoveToHeight(m_elevator, ElevatorConstants.kMinHeight));
```
AnthonyFurman, line 41: these should be in the test controls.

src/main/java/frc/robot/commands/elevator/ManualElevatorControl.java
```diff
@@ -0,0 +1,56 @@
...
+    addRequirements(m_elevator);
+  }
+  @Override
+  public void initialize() {
+    m_elevator.setEnabled(false); 
+
+  }
+  @Override
+  public void execute() {
+    if ((m_operator.getClampedThrottleValue() > 0) || (m_operator.getClampedThrottleValue() < 0)){
```
AnthonyFurman, line 31: shouldn't this use the manual controller?

PandaNinja10:
Closeing as future PRs will be though a superstructure intigration branch

PandaNinja10:
nvm probobly better to have smaller PRs so make sure this PR only has Elevator stuff and all the related things

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -53,9 +56,10 @@ public class RobotContainer {
   private final ShuffleboardTab m_testTab = Shuffleboard.getTab("Test");
 
   // The robot's subsystems are defined here...
-  private final Drivetrain m_drive;
-  private final FourBarArm m_arm;
-  private final Intake m_intake;
+  private Drivetrain m_drive;
+  private FourBarArm m_arm;
+  private Intake m_intake;
+  private Elevator m_elevator;
```
AnthonyFurman, line 63: Not relevant to PR and shouldn't be done

src/main/java/frc/robot/commands/elevator/DoNothing.java
```diff
@@ -1,4 +1,4 @@
-package frc.robot.commands;
+package frc.robot.commands.elevator;
 
 import edu.wpi.first.wpilibj2.command.InstantCommand;
```
AnthonyFurman, line 3: You accidentally moved this file,. please move it back

src/main/java/frc/robot/constants/ElevatorConstants.java
```diff
@@ -0,0 +1,40 @@
...
+  public static final double kP = 0.1;
+  public static final double kI = 0.0;
+  public static final double kD = 0.0;
+
+  //these feedforward values were calculated using : https://www.reca.lc/linear
+  public static final double kS = 0.0000001; //FIXME: Not sure what to put for this feedforward value. 
+  public static final double kG = 0.28; 
+  public static final double kV = 10.72; 
+  public static final double kA = 0.05; 
+
```
AnthonyFurman, line 21: How did you get these values ? Reca.lc?

src/main/java/frc/robot/controls/Driver.java
```diff
@@ -0,0 +1,20 @@
...
+  private static GameController driver = new GameController(OIConstants.kDriverJoy);
+
+  /**
+   * Configures all the driver controls, which are the default controls for the robot.
+   */
+  public static void configureControls(Drivetrain drive) {
+    
+    // example button binding implementation
+    driver.get(Button.A).onTrue(new DoNothing());
+  }
```
AnthonyFurman, line 19: Remove file

src/main/java/frc/robot/subsystems/Elevator.java
```diff
@@ -0,0 +1,169 @@
...
+     * Thus set the setpoint to 0, then atSetpoint() will be triggered, causing the 
+     * ResetEncoderAtBottom() command to end
+     */
+    
+    if(m_enabled) {
+      double pid = m_elevatorPID.calculate(getHeight());
+      double ff = m_elevatorFF.calculate(ElevatorConstants.kVelocity, ElevatorConstants.kAccel); 
+
+      double pidClamped = MathUtil.clamp(pid, -ElevatorConstants.kPowerLimit, ElevatorConstants.kPowerLimit);
+      double finalMotorPower = pidClamped + ff;
```
AnthonyFurman, line 74: The pid and ff should be clamped together

src/main/java/frc/robot/subsystems/Elevator.java
```diff
@@ -0,0 +1,169 @@
...
+  public void periodic() {
+    /**
+     * If we hit the bottom limit switch, then we must be at the zero position.
+     * Thus set the setpoint to 0, then atSetpoint() will be triggered, causing the 
+     * ResetEncoderAtBottom() command to end
+     */
+    
+    if(m_enabled) {
+      double pid = m_elevatorPID.calculate(getHeight());
+      double ff = m_elevatorFF.calculate(ElevatorConstants.kVelocity, ElevatorConstants.kAccel);
```
AnthonyFurman, line 71: Im not sure you need a ff of velocity and acceleration, i was just thinking gravity (that might increase once it passes a certain stage), for now now I would remove this. Additionally this is not how you would calculate the ff, the velocity and acceleration are not constant.

src/main/java/frc/robot/subsystems/Elevator.java
```diff
@@ -0,0 +1,169 @@
...
+  public void close() {
+    // close the ports
+    //TODO: Is there any way to close the ports of the limit switches attached to the motor? 
+    m_topLimitSwitch.close(); 
+    m_bottomLimitSwitch.close(); 
+    m_absEncoder.close();
+  }
+
+  public void set(double power) {
+    if((m_motor.get() > 0 && isBottom()) || (m_motor.get() < 0 && isTop())){
```
AnthonyFurman, line 91: m_motor.get() would be the last motor power, you should check the set motor power which is called `power` here, the parameter. Also, if neither of the limit switches are triggered, this won't let you set the motors

src/main/java/frc/robot/subsystems/Elevator.java
```diff
@@ -0,0 +1,169 @@
...
+
+  // public boolean getTopLimitSwitch(){
+  //   if(m_motor.getSensorCollection().isFwdLimitSwitchClosed() == 0){
+  //    return true; 
+  //   }
+  //    return false; 
+  // }
+
+  public boolean isBottom(){
+    if(m_bottomLimitSwitch.get() == false){
```
AnthonyFurman, line 122: You can just do `return limitSwitch.get()`

lakshag32:
This branch is outdated so I have closed the pull request for it.

PR #17, opened 2023-02-12
Vision

Kyle-Eldridge:
## Overview
This PR will merge the Vision branch with the main branch.
The Vision branch has everything related to the vision system on the robot, which currently uses April tags.

## Design Doc
https://docs.google.com/document/d/1UfKGZ_5jCUP1P7YjEF6BluMhBr28Fxi27rlz2LOM77A/edit
For testing, we had one camera in the front and one in the back, but we are going to have 2 in the front for the actual robot. The only thing that needs to be changed is the VisionConstants file. We also do not currently have logging other than printing because we were using a different repository. Because the classbot broke, we were not able to test the alignment test, but it exists. We also currently do not have anything for retroreflective tape, only April tags. Other than that, we have everything in the design doc, including 2 working cameras.

## Tests Ran
We have three tests: drive and print distance from encoders and vision, drive 3 meters and then go back near the starting pose, and align to a specified angle. For the first one, we got that the difference between the encoders and vision was consistently about 24-25%. We were using a classbot, so the encoders might not have been accurate, and the camera probably needed to be recalibrated. We ran the second test, bit I don't remember the exact results, and the robot broke before we could run it more. The third test broke the robot. I think I fixed the test, but we were not able to test it.

## Additional Notes
The robot is able to use two cameras to find April tags, including more than one at a time. The pose estimation should be better on the actual robot.


pathweaver.json
```diff
@@ -0,0 +1,8 @@
+{
+  "lengthUnit": "FOOT",
+  "exportUnit": "Always Meters",
+  "maxVelocity": 10.0,
+  "maxAcceleration": 60.0,
+  "trackWidth": 2.0,
+  "gameName": "Infinite Recharge"
```
AnthonyFurman, line 7: We are not using pathweaver, also we don't use meters nor are we playing infinite recharge.

src/main/java/frc/robot/Robot.java
```diff
@@ -4,12 +4,22 @@
...
 import edu.wpi.first.wpilibj.smartdashboard.SmartDashboard;
 import edu.wpi.first.wpilibj2.command.Command;
 import edu.wpi.first.wpilibj2.command.CommandScheduler;
+import frc.robot.RobotContainer.Teams;
 import frc.robot.util.LogManager;
+import frc.robot.util.Node;
+import frc.robot.util.PathGroupLoader;
+import frc.robot.util.Vision;
+import lib.controllers.GameController.DPad;
+// import frc.robot.RobotContainer.Teams;
```
AnthonyFurman, line 22: really minor but: this import is commented out then included above?

src/main/java/frc/robot/Robot.java
```diff
@@ -19,6 +29,11 @@
  */
 public class Robot extends TimedRobot {
   private Command m_autoCommand;
+  
+  
+  // Where the robot will score.
```
AnthonyFurman, line 34: what does this mean?

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -23,11 +30,38 @@
...
+
+  // 2D arrays of nodes. blueNodes[3][1] will return the top row cone node on the far left side (from the perspective of the driver)
+  public final static Node[][] blueNodes = new Node[4][];
+  public final static Node[][] redNodes = new Node[4][];
+
+
+
+  // Possible teams
+  public static enum Teams {BLUE, RED};
+  public static Teams team;
```
AnthonyFurman, line 64: You can use Alliance.Red or Alliance.Blue, that is a already included WPI class that the driverstation class uses.

src/main/java/frc/robot/Robot.java
```diff
@@ -77,6 +98,37 @@ public void robotPeriodic() {
...
   }
 
+    /**
+   * Method to store DPad values and use them to set selectedNode
+   * Down clears the array
+   * Left is 1, up is 2, and right is 3 for selection
+   * For example, up right left will select the center grid, top row, and left spot.
+   * @param direction = Which DPad button is pressed
+   */
+  public static void DPadPress(DPad direction) {
```
AnthonyFurman, line 108: selectTime and selectValues are not very clear, additionally, this should be in Driver.java or Operator.java

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -23,11 +30,38 @@
  * subsystems, commands, and trigger mappings) should be declared here.
  */
 public class RobotContainer {
+
   // The robot's subsystems are defined here...
   private final Drivetrain m_drive = new Drivetrain();
 
   // Shuffleboard stuff
   SendableChooser<Command> m_autoCommand = new SendableChooser<>();
+  SendableChooser<Teams> m_teamChooser = new SendableChooser<>();
```
AnthonyFurman, line 39: I think you can use Driverstation.getAlliance() or something like that to get the alliance. That will simplify a bit.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -23,11 +30,38 @@
...
+
+  // How much time it should take (in frames)
+  public final static double selectTimeAmount=100;
+
+  // Array of april tags. The index of the april tag in the array is equal to its id, and aprilTags[0] is null.
+  public final static Pose3d[] aprilTags = new Pose3d[9];
+
+  // 2D arrays of nodes. blueNodes[3][1] will return the top row cone node on the far left side (from the perspective of the driver)
+  public final static Node[][] blueNodes = new Node[4][];
+  public final static Node[][] redNodes = new Node[4][];
```
AnthonyFurman, line 58: shouldn't the second dimension size be defined?

src/main/java/frc/robot/commands/Align.java
```diff
@@ -0,0 +1,91 @@
...
+import edu.wpi.first.wpilibj2.command.ParallelCommandGroup;
+import edu.wpi.first.wpilibj2.command.SequentialCommandGroup;
+import frc.robot.Robot;
+import frc.robot.subsystems.Drivetrain;
+import frc.robot.util.Vision;
+
+public class Align extends CommandBase{
+  private Drivetrain m_drive;
+  private double setpoint;
+  private double angle;
```
AnthonyFurman, line 21: m_ for these variables

src/main/java/frc/robot/commands/Align.java
```diff
@@ -0,0 +1,91 @@
...
+
+  /**
+   * Turns the robot
+   * PID broke the classbot, so this doesn't use it
+   */
+  @Override
+  public void execute(){
+    angle=getAngle();
+    double a = angle;
+    if(a<0&&setpoint>0){
```
AnthonyFurman, line 60: Use MathUtil.angleModulus for this. The way you programmed it doesn't guarantee the angle will be the closest.

src/main/java/frc/robot/commands/MoveArm.java
```diff
@@ -0,0 +1,19 @@
+package frc.robot.commands;
+
+import edu.wpi.first.math.geometry.Pose2d;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import edu.wpi.first.wpilibj2.command.ParallelCommandGroup;
+import edu.wpi.first.wpilibj2.command.SequentialCommandGroup;
+import frc.robot.Robot;
+import frc.robot.subsystems.Drivetrain;
+
+public class MoveArm extends CommandBase{
```
AnthonyFurman, line 10: What is this for?

src/main/java/frc/robot/commands/MoveToPose.java
```diff
@@ -0,0 +1,20 @@
+package frc.robot.commands;
+
+import edu.wpi.first.math.geometry.Pose2d;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import edu.wpi.first.wpilibj2.command.ParallelCommandGroup;
+import edu.wpi.first.wpilibj2.command.SequentialCommandGroup;
+import frc.robot.Robot;
+import frc.robot.subsystems.Drivetrain;
+
+public class MoveToPose extends CommandBase{
```
AnthonyFurman, line 10: This is being done with the swerve team, please don't add unimplemented commands.

src/main/java/frc/robot/commands/ScoreCommand.java
```diff
@@ -0,0 +1,34 @@
...
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import edu.wpi.first.wpilibj2.command.ParallelCommandGroup;
+import edu.wpi.first.wpilibj2.command.SequentialCommandGroup;
+import frc.robot.Robot;
+import frc.robot.RobotContainer;
+import frc.robot.RobotContainer.Teams;
+import frc.robot.subsystems.Drivetrain;
+import frc.robot.util.Node;
+
+public class ScoreCommand extends SequentialCommandGroup{
```
AnthonyFurman, line 14: It's not really needed to do this, I think it's mostly clear how the commands will go

src/main/java/frc/robot/commands/TestVision.java
```diff
@@ -0,0 +1,104 @@
...
+  private Pose2d currentPose = null;
+  private int endCounter = 0;
+  private int printCounter=0;
+  private double m_speed;
+
+  //How many frames it has to not see anything to end the command
+  private final int endDelay = 5;
+
+  //How many frames it will wait between prints
+  private final int printDelay = 50;
```
AnthonyFurman, line 30: m_ for private variables

src/main/java/frc/robot/commands/TestVision.java
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  private double getDist(){
+    // return 0;
+    return m_drive.getLeftDistance()/2+m_drive.getRightDistance()/2;
+  }
+
+  private Pose2d getPose(){
+    Optional<Pair<Pose3d, Double>> p = Vision.getEstimatedGlobalPose(currentPose==null?m_drive.getPose():currentPose);
+    if(p.isPresent() && p.get().getFirst() != null && p.get().getSecond() != null && p.get().getFirst().getX() > -10000 && p.get().getSecond() >= 0){
```
AnthonyFurman, line 50: This code seems pretty similar to code before, checking for null stuff can be handled in vision.java and just be part of getEstimatedGlobalPose

src/main/java/frc/robot/commands/TestVision.java
```diff
@@ -0,0 +1,104 @@
...
+      endCounter++;
+    }else{
+      endCounter = 0;
+      printCounter++;
+      currentPose = getPose();
+      encoderPosition = getDist();
+      if(printCounter%printDelay==0){
+        double dist1 = Math.abs(encoderPosition-encoderStart);
+        double dist2 = Math.sqrt(Math.pow(currentPose.getX()-startPose.getX(), 2) + Math.pow(currentPose.getY()-startPose.getY(), 2));
+        System.out.printf("\nEncoder distance: %.4f\nVision distance: %.4f\nDifference: %.4f\nPercent difference: %.4f%%\n", dist1, dist2, dist2-dist1, (dist2-dist1)/dist1*100);
```
AnthonyFurman, line 82: Some of this will need to change with the swerve drivetrain. That PR will be merged soon, so be ready for that. I also think it should indicate if the %error is within a certain bound. You said on the classbot the error was ~25%, that seems too high for me.

src/main/java/frc/robot/commands/TestVision2.java
```diff
@@ -0,0 +1,123 @@
...
+    m_speed=speed;
+    distanceToMove=distance;
+  }
+
+  private double getDist(){
+    // return 0;
+    return m_drive.getLeftDistance()/2+m_drive.getRightDistance()/2;
+  }
+
+  private Pose2d getPose(){
```
AnthonyFurman, line 50: like before, this should be a function in Vision.java

src/main/java/frc/robot/constants/Constants.java
```diff
@@ -38,4 +38,8 @@ public final class Constants {
    */
   @Deprecated
   public static final FalconConstants falcon = new FalconConstants();
+  @Deprecated
+  public static final VisionConstants vision = new VisionConstants();
+  @Deprecated
+  public static final FieldConstants field = new FieldConstants();
```
AnthonyFurman, line 44: you don't need to add this

src/main/java/frc/robot/constants/VisionConstants.java
```diff
@@ -0,0 +1,43 @@
...
+  public static final String kCameraName1 = "Camera_1";
+  public static final String kCameraName2 = "Camera_2";
+
+  //Distance and rotation from the center of the robot to the cameras
+  public static final Transform3d kRobotToCamera1 = new Transform3d(new Translation3d(Units.inchesToMeters(10), 0, Units.inchesToMeters(5)), new Rotation3d(0, 0, Math.PI));
+  public static final Transform3d kRobotToCamera2 = new Transform3d(new Translation3d(Units.inchesToMeters(-10), 0, Units.inchesToMeters(5)), new Rotation3d(0, 0, 0));
+  
+  //Currently not used. Do not change.
+  public static final double kCamera1Rotation=0;
+  public static final double kCamera2Rotation=0;
```
AnthonyFurman, line 37: If it's not used, what is it for?

src/main/java/frc/robot/controls/Operator.java
```diff
@@ -1,17 +1,22 @@
...
-import frc.robot.constants.OIConstants;
+import frc.robot.constants.Constants;
+import frc.robot.subsystems.Drivetrain;
 import lib.controllers.GameController;
 import lib.controllers.GameController.Button;
+import lib.controllers.GameController.DPad;
 
 public class Operator {
-  private static GameController operator = new GameController(OIConstants.kOperatorJoy);
+  private static GameController operator = new GameController(Constants.oi.kOperatorJoy);
```
AnthonyFurman, line 13: Use OIConstants.XYZ, not Constants.oi.XYZ

src/main/java/frc/robot/util/Node.java
```diff
@@ -0,0 +1,90 @@
...
+    public final Types type;
+
+    /**
+     * Creates a new Node object
+     * @param team
+     *  The node's color
+     * @param row
+     *  Which row it's in (1 = bottom, 2 = middle, 3 = top)
+     * @param x
+     *  X position of the node from the alliance area (1-9)
```
AnthonyFurman, line 59: Instead of "x", this should probably be called gridColumn or similar

src/main/java/frc/robot/util/Node.java
```diff
@@ -0,0 +1,90 @@
...
+     */
+    public Node(Teams team, int row, int x){
+        this.team=team;
+        this.row=row;
+        this.x=x;
+        type = row==1?Types.HYBRID:x%3==2?Types.CUBE:Types.CONE;
+
+        Pose3d tag = RobotContainer.aprilTags[(x-1)/3+(team==Teams.RED?1:6)];
+        double x2;
+        double z;
```
AnthonyFurman, line 69: could have better names

src/main/java/frc/robot/util/Vision.java
```diff
@@ -0,0 +1,176 @@
...
+
+
+public class Vision {
+  private static RobotPoseEstimator robotPoseEstimator;
+  // private static RobotPoseEstimator robotPoseEstimator2;
+  private static AprilTagFieldLayout aprilTagFieldLayout;
+ 
+
+  /**
+   * Calls the other setup function
```
AnthonyFurman, line 55: javadoc comments shouldn't explain how the function does something, but just what the function does.

AnthonyFurman: For example, if I hover over this function, I will just see "Calls the other setup function" which tells me nothing

src/main/java/frc/robot/util/Vision.java
```diff
@@ -0,0 +1,176 @@
...
+      camList = new ArrayList<Pair<PhotonCamera, Transform3d>>(List.of(
+        new Pair<PhotonCamera, Transform3d>(camera1, VisionConstants.kRobotToCamera1)
+      ));
+    }
+    getTagFieldLayout();
+    aprilTagFieldLayout.setOrigin(AprilTagFieldLayout.OriginPosition.kBlueAllianceWallRightSide);
+    
+    robotPoseEstimator = new RobotPoseEstimator(aprilTagFieldLayout, PoseStrategy.CLOSEST_TO_REFERENCE_POSE, camList);
+
+    if(VisionConstants.k2Cameras){
```
AnthonyFurman, line 83: consider removing this

src/main/java/frc/robot/util/Vision.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+  }
+
+  public static AprilTagFieldLayout getTagFieldLayout() {
+    // AprilTagFields.loadAprilTagLayoutField();
+    try {
+      //TODO: Fix this
+      // aprilTagFieldLayout = AprilTagFieldLayout.loadFromResource(AprilTagFields.k2023ChargedUp.m_resourceFile);
+      aprilTagFieldLayout = AprilTagFieldLayout.loadFromResource(AprilTagFields.k2022RapidReact.m_resourceFile);
```
AnthonyFurman, line 97: I saw your writing that this doesn't work, why doesn't it? Is the file not there yet?

src/main/java/frc/robot/util/Vision.java
```diff
@@ -0,0 +1,176 @@
...
+    //     return new Pair<Pose2d, Double>(null, 0.0);
+    // }
+}
+
+/**
+ * Not used
+ * @param prevEstimatedRobotPose Doesn't do anything
+ * @return null
+ */
+public static Optional<Pair<Pose3d,Double>> getEstimatedGlobalPose2(Pose2d prevEstimatedRobotPose) {
```
AnthonyFurman, line 155: consider removing this

src/main/java/frc/robot/util/Vision.java
GLRoylance: The April Tag locations in the file are for RapidReact (i.e., for testing on last year's field). Same for tag locations in kDefault.

src/main/java/frc/robot/RobotContainer.java
Kyle-Eldridge: No. 2D arrays can have arrays of different lengths inside of them. The size is defined later, when those arrays are filled.

src/main/java/frc/robot/commands/MoveArm.java
Kyle-Eldridge: Basic command structure for how to use selectedNode.

src/main/java/frc/robot/commands/MoveToPose.java
Kyle-Eldridge: I just wanted to put something there, for how to use selectedNode.

src/main/java/frc/robot/commands/TestVision.java
Kyle-Eldridge: Maybe, but I might want other information, or I might not want it to return null somewhere else.

src/main/java/frc/robot/commands/TestVision.java
Kyle-Eldridge: It was, but I don't know how accurate it should be on the actual robot.

src/main/java/frc/robot/util/Vision.java
GLRoylance: I put the AprilTag locations into a new pull request. The April Tag locations are part of the PowerPlay field.

src/main/java/frc/robot/commands/Align.java
AnthonyFurman: If the angle is only those values it will technically work but it can introduce a bug if the angle is a different range. Additionally, it is clearer and uses 4 less lines. 
To do this, just do this: MathUtil.angleModulus(double radians). The parameter is the angle in radians, it will output it in radians from -pi to pi.

src/main/java/frc/robot/commands/MoveArm.java
AnthonyFurman: We don't need to write command structure like this

src/main/java/frc/robot/commands/MoveToPose.java
AnthonyFurman: I would add that to the comments or design doc. It just seems inefficient to create three classes to do nothing but document a function. selected node should probably be a function btw.

src/main/java/frc/robot/commands/TestVision.java
AnthonyFurman: At least <10%

AnthonyFurman:
@Kyle-Eldridge do not mark my comments as resolved unless you committed something to fix them...

src/main/java/frc/robot/util/Vision.java
Kyle-Eldridge: If we want to remove one camera for any reason, we only need to change it in the constants file. There is also no reason to remove this.

src/main/java/frc/robot/util/Vision.java
Kyle-Eldridge: I don't know why. It can't find k2023ChargedUp. I think the file needs to be updated, but I can't find it anywhere.

src/main/java/frc/robot/util/Vision.java
Kyle-Eldridge: This is for a version that doesn't work. If we decide to go back to it, I don't want to rewrite it.

GLRoylance:
See Richie's comment and the change to build.gradle in PR #18.

The simple version is to use kDefault; it has the ChargedUp field in
2023.2.1.

src/main/java/frc/robot/commands/MoveToPose.java
Kyle-Eldridge: It can be deleted when someone has a better version.

src/main/java/frc/robot/commands/TestVision.java
Kyle-Eldridge: I created a new function in Vision.java. I can't change the other one because the drivetrain needs the latency.

src/main/java/frc/robot/commands/TestVision.java
Kyle-Eldridge: Probably less if the encoders are good, but I don't know. <5% might be good, but it might need to be even more accurate.

src/main/java/frc/robot/commands/Align.java
Kyle-Eldridge: This is already always between -pi and pi. This code is so it knows -179 degrees is close to 180 degrees, and it should not turn 359 degrees to get to it.

Kyle-Eldridge:
The error where it couldn't find the constant for the April tag field layout is fixed. Ignore the last part of the PR that mentioned line 96.

build.gradle
```diff
@@ -1,6 +1,6 @@
 plugins {
     id "java"
-    id "edu.wpi.first.GradleRIO" version "2023.1.1"
+    id "edu.wpi.first.GradleRIO" version "2023.3.2"
     id "com.peterabeles.gversion" version "1.10"
```
GLRoylance, line 4: GradleRIO should be updated in the main branch quickly. Several Pull Requests have updated it to 2023.2.1. The change is appropriate for more than just vision, so it should be a separate request.

https://plugins.gradle.org/plugin/edu.wpi.first.GradleRIO shows that your 2023.3.2 is the latest.

Make a separate Pull Request that is just this file. That PR should be approved quickly.

vendordeps/NavX.json
```diff
@@ -0,0 +1,39 @@
+{
+    "fileName": "NavX.json",
+    "name": "KauaiLabs_navX_FRC",
+    "version": "2023.0.3",
```
GLRoylance, line 4: I would put this whole file in a separate Pull Request.

vendordeps/photonlib.json
```diff
@@ -0,0 +1,41 @@
+{
+    "fileName": "photonlib.json",
+    "name": "photonlib",
+    "version": "v2023.1.2",
+    "uuid": "515fe07e-bfc6-11fa-b3de-0242ac130004 ",
```
GLRoylance, line 5: This whole file could be put in its own PR or in the NavX PR.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -24,12 +32,28 @@
...
 
   // Shuffleboard stuff
   SendableChooser<Command> m_autoCommand = new SendableChooser<>();
+  SendableChooser<Alliance> m_allianceChooser = new SendableChooser<>();
+
+  // Where the robot will score.
+  public static Node selectedNode = null;
+
+  // Array of april tags. The index of the april tag in the array is equal to its id, and aprilTags[0] is null.
+  public final static Pose3d[] aprilTags = new Pose3d[9];
```
GLRoylance, line 48: This code looks very suspect. It suggests that the Pose3d of tags 1-8 is (0, 0, 0) (0, 0, 0).

Furthermore, the array is not needed -- just call the Field method that provides Optional<Pose3d>.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -40,16 +64,36 @@ public RobotContainer() {
...
+    Driver.configureControls(m_drive, m_arm);
+    Operator.configureControls(m_drive, m_arm);
     TestControls.configureControls(m_drive);
 
+    Vision.setup(m_drive);
+
+    // Puts April tags in array
+    for(int i = 1; i <= 8; i++){
+      aprilTags[i] = Vision.getTagPose(i);
+    }
```
GLRoylance, line 76: Now the values are changed to actual Pose3D values, but why is aprilTags needed when there is Vision.getTagPose(i)?

src/main/java/frc/robot/util/Node.java
```diff
@@ -0,0 +1,67 @@
...
+     *  Grid column from left to right (1-9)
+     */
+    public Node(Alliance alliance, int row, int x){
+        this.alliance=alliance;
+        this.row=row;
+        this.x=x;
+        type = row==1?Types.HYBRID:x%3==2?Types.CUBE:Types.CONE;
+
+        // Which April tag this node is closest to
+        Pose3d tag = RobotContainer.aprilTags[(x-1)/3+(alliance==Alliance.Red?1:6)];
```
GLRoylance, line 39: Why not just save the tag number? We do not need to store Pose3D values in many places.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -61,13 +105,28 @@ public RobotContainer() {
   public Command getAutonomousCommand() {
     return m_autoCommand.getSelected();
   }
+  public void allianceChooserUpdate() {
+    alliance = DriverStation.getAlliance();
+    // alliance = m_allianceChooser.getSelected();
+    // Shuffleboard.getTab("Main").add("Alliance Chooser", m_allianceChooser);
+  }
```
GLRoylance, line 112: This practice seems odd.

You are calling this in disabledPeriodic. If the Alliance is needed, why not just call the DriverStation method when it is needed?

src/main/java/frc/robot/Robot.java
```diff
@@ -85,7 +95,10 @@ public void disabledInit() {
 
   /** This function is called periodically when the robot is disabled */
   @Override
-  public void disabledPeriodic() {}
+  public void disabledPeriodic() {
+    m_robotContainer.allianceChooserUpdate();
+    m_autoCommand = m_robotContainer.getAutonomousCommand(); // update the auto command before auto starts
```
GLRoylance, line 100: Delete these. They do not need to be done every 20 milliseconds.

The first sets the Alliance. Call the DriverStation method when it is needed.

The autonomous command selection is done (once) in autonomousInit().

build.gradle
AnthonyFurman: I've updated main to the latest.

src/main/java/frc/robot/constants/Constants.java
```diff
@@ -15,27 +15,31 @@ public final class Constants {
    * Operator Input Constants
    * @deprecated Instead of frc.robot.constants.Constants.oi use frc.robot.constants.OIConstants
    */
-  @Deprecated
-  public static final OIConstants oi = new OIConstants();
```
AnthonyFurman, line 19: You don't need to change this file

src/main/java/frc/robot/constants/VisionConstants.java
AnthonyFurman: @Kyle-Eldridge don't resolve my comments without addressing them. Your PR will not be approved.

AnthonyFurman: I assume this is for if a camera is rotated though right?

src/main/java/frc/robot/util/Node.java
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * Creates a new Node object
+     * @param team
+     *  The node's color
+     * @param row
+     *  Which row it's in (1 = bottom, 2 = middle, 3 = top)
+     * @param x
+     *  Grid column from left to right (1-9)
+     */
+    public Node(Alliance alliance, int row, int x){
```
AnthonyFurman, line 32: x should be called column or something similiar

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -24,12 +32,28 @@
...
  */
 public class RobotContainer {
+
   // The robot's subsystems are defined here...
   private final Drivetrain m_drive = new Drivetrain();
   private final FourBarArm m_arm = new FourBarArm();
 
   // Shuffleboard stuff
   SendableChooser<Command> m_autoCommand = new SendableChooser<>();
+  SendableChooser<Alliance> m_allianceChooser = new SendableChooser<>();
```
AnthonyFurman, line 42: I might have already said this, I'm not sure where the comment went, but just use Driverstation.getAlliance()

src/main/java/frc/robot/RobotContainer.java
Kyle-Eldridge: This array is filled later, after Vision.setup() is called. It just makes it easier to access an April tag. aprilTags[1] is a lot easier than Vision.getTagPose(1). It also avoids any possible error where reference variables aren't equal.

src/main/java/frc/robot/RobotContainer.java
Kyle-Eldridge: Read previous comment.

src/main/java/frc/robot/util/Node.java
Kyle-Eldridge: Again, it avoids any possible error if we ever want to do something like aprilTags[1]==Vision.getTagPose(1). If they're all from the array, they will always reference the same variable.

src/main/java/frc/robot/RobotContainer.java
Kyle-Eldridge: I changed it to not have alliance stored in RobotContainer.

src/main/java/frc/robot/constants/Constants.java
Kyle-Eldridge: Maybe, but I already did We aren't supposed to use Constants.oi anyway, so there is no reason to add it back.

src/main/java/frc/robot/commands/DefaultDriveCommand.java
```diff
@@ -28,7 +30,7 @@ public void execute() {
     double ySpeed = m_driver.getSideTranslation();
     double rot = m_driver.getRotation();
 
-    m_drive.drive(xSpeed, ySpeed, rot, true);
+    m_drive.drive(DriverStation.getAlliance()==Alliance.Blue?xSpeed:-xSpeed, DriverStation.getAlliance()==Alliance.Blue?ySpeed:-ySpeed, rot, true);
```
AnthonyFurman, line 33: This wouldn't be necessary because the pigeon would be reset to be facing the other direction when you are on the other alliance. It's already kind of done by the constant kStartingHeading, or something like that.
But beyond that, I would like to note two things:
1) I would prefer this be done somehow in the init()
2) This code has nothing to do with vision, so it shouldn't be on this branch/PR

src/main/java/frc/robot/commands/MoveToPose.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Default drive command. Drives robot using driver controls.
+ */
+public class MoveToPose extends CommandBase {
+
+  private final Drivetrain m_drive;
+  private Pose2d m_pose;
+  
+  public MoveToPose(Drivetrain drive, Pose2d pose) {
```
AnthonyFurman, line 15: This is already a command. It also factors in a relative transformation, to move a distance from the robot, rather than a global pose. I'm also not sure it's necessary in the Vision PR.

src/main/java/frc/robot/commands/PathPlannerCommand.java
```diff
@@ -0,0 +1,54 @@
...
+import com.pathplanner.lib.commands.PPSwerveControllerCommand;
+
+import edu.wpi.first.wpilibj2.command.InstantCommand;
+import edu.wpi.first.wpilibj2.command.PrintCommand;
+import edu.wpi.first.wpilibj2.command.SequentialCommandGroup;
+import frc.robot.constants.AutoConstants;
+import frc.robot.subsystems.Drivetrain;
+import frc.robot.util.PathGroupLoader;
+
+public class PathPlannerCommand extends SequentialCommandGroup {
```
AnthonyFurman, line 20: I'm pretty sure this is also on a different branch. You're being very ineffiecent if you're working on code we later have to delete because someone else already did it. I understand you need pathplanner stuff for the grid alignment, but the grid alignment should be on a different branch. Make this branch just basic vision, that supports getPose2d(). After that is approved you can have grid alignment, etc.

AnthonyFurman:
@Kyle-Eldridge @richtan @PandaNinja10 this PR is too big. Could we make a simple PR of the fewest # of possible features? You might want to make a new branch for that. Ideally it would just be adding the vision support that has a getPose2d() function. That way, that can be merged and you can start working on the other features, on different branches.
Not only does this PR have too many vision features, there are also unrelated features. For example, Arm.java and the constants. That's a great change, but instead of being on a different PR that would have been approved immediately, it is now held up in this PR and probably won't be approved for a while. In other words, if we make more, smaller PRs, they get approved faster, things get fixed, and kyle gets to work on more features. It also makes it a lot easier to review :).

I think this is a failure on my part for not keeping better track of this and letting this happen again. We should have been more clear with when to make a branch from the start. My apologies. For now the best we can do is split a good amount of the features to different branches.

src/main/java/frc/robot/commands/DefaultDriveCommand.java
Kyle-Eldridge: I added it because the controls were inverted for testing, but I'll change it back.

src/main/java/frc/robot/commands/MoveToPose.java
Kyle-Eldridge: This is for moving to a global pose, in front of the grids for scoring (using vision). There was no other command on the branch that does the same thing. I also changed it since then to only move to the selected spot.

src/main/java/frc/robot/commands/PathPlannerCommand.java
Kyle-Eldridge: I copied this form RTVision, and we can replace it with a newer version. I can change it to not align to grids yet.

Kyle-Eldridge:
I think this PR is done. We can fix the grid alignment in a separate PR.

Kyle-Eldridge:
Vision works now. It isn't updating the robot pose, but I think that's because vision was returning a value over 15 meters away from the default position, and it doesn't trust the vision enough. Other than that, it is able to return a pose, and the distance between the start and end poses is relatively accurate (I wasn't able to measure it exactly). I think this is good enough to merge with main, and we can fix it later if we have to. The encoders are not accurate (75% different from vision), so the test results don't give me any useful information.

@AnthonyFurman  @PandaNinja10

src/main/java/frc/robot/commands/AlignToRow.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.List;
+
+import com.pathplanner.lib.PathPoint;
+
+import edu.wpi.first.wpilibj2.command.SequentialCommandGroup;
+import frc.robot.controls.Operator;
+import frc.robot.subsystems.Drivetrain;
+
+/**
+ * Go to 1-9 row in grid area
```
PandaNinja10, line 13: column

src/main/java/frc/robot/commands/vision/TestVision.java
```diff
@@ -0,0 +1,96 @@
...
+    m_vision=vision;
+  }
+
+  private double getDist(){
+    return 
+      m_drive.m_modules[0].getDrivePosition()/4+
+      m_drive.m_modules[1].getDrivePosition()/4+
+      m_drive.m_modules[2].getDrivePosition()/4+
+      m_drive.m_modules[3].getDrivePosition()/4;
+  }
```
PandaNinja10, line 44: use the odometry pose rather than individual module encoders

src/main/java/frc/robot/commands/vision/TestVision.java
```diff
@@ -0,0 +1,96 @@
...
+      m_endCounter++;
+    }else{
+      m_endCounter = 0;
+      m_printCounter++;
+      m_currentPose = pose;
+      m_encoderPosition = getDist();
+      if(m_printCounter%printDelay==0){
+        double dist1 = Math.abs(m_encoderPosition-m_encoderStart);
+        double dist2 = Math.sqrt(Math.pow(m_currentPose.getX()-m_startPose.getX(), 2) + Math.pow(m_currentPose.getY()-m_startPose.getY(), 2));
+        System.out.printf("\nEncoder distance: %.4f\nVision distance: %.4f\nDifference: %.4f\nPercent difference: %.4f%%\n", dist1, dist2, dist2-dist1, (dist2-dist1)/dist1*100);
```
PandaNinja10, line 74: calculation of dist 2 and printing are too long for one line please split this acorss lines (also there is a hypotonose function in Util)

src/main/java/frc/robot/commands/vision/TestVision.java
```diff
@@ -0,0 +1,96 @@
...
+
+  //How many frames it will wait between prints
+  private static final int printDelay = 50;
+
+  /**
+   * A command that moves and prints out distances
+   * @param speed What speed to move at
+   * @param drive The drivetrain
+   * @param vision The vision
+   */
```
PandaNinja10, line 30: maybe add that it moves until it dosen't get see an april tag

src/main/java/frc/robot/commands/vision/TestVision.java
PandaNinja10, line 74: consider useing shuffleboard. Not mandatory but might be more convinient

src/main/java/frc/robot/commands/vision/TestVision2.java
```diff
@@ -0,0 +1,113 @@
...
+    m_turned=false;
+    m_endCounter=0;
+    m_closest=100;
+  }
+
+  /**
+   * Moves the robot
+   * If it has moves farther than the specified distance, it turns around
+   * It stops when it can't see an April tag or it has passed its starting position
+   */
```
PandaNinja10, line 64: better grammer and a better explination would be good for people to understand what is going on. Similarly please have better name than vison test 2 and same for the original vision test.

src/main/java/frc/robot/commands/vision/TestVision2.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+  }
+
+  /**
+   * Stops the robot and prints the distance it traveled
+   * @param interrupted If the command was interrupted
+   */
+  @Override
+  public void end(boolean interrupted){
+    System.out.printf("\nVision distance: %.4f\nEncoder distance: %.4f\n", Math.sqrt(Math.pow(m_currentPose.getX()-m_startPose.getX(), 2) + Math.pow(m_currentPose.getY()-m_startPose.getY(), 2)), Math.abs(m_encoderPosition-m_encoderStart));
```
PandaNinja10, line 101: this line of code is too long please split it over mulitple lines. Personaly I think multiple printf statments would be best.

src/main/java/frc/robot/commands/vision/TestVision2.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  /**
+   * Initializes the command
+   */
+  @Override
+  public void initialize(){
+    m_encoderStart=getDist();
+    m_startPose=m_vision.getPose2d(m_currentPose, m_drive.getPose());
+    m_direction=1;
```
PandaNinja10: why dose  m_direction exist when you have m_turned or vice versa. you even use m_direction to check if it turned for one rather than useing m_turrned.

src/main/java/frc/robot/commands/vision/TestVision2.java
```diff
@@ -0,0 +1,113 @@
...
+    m_drive.drive(m_direction*m_speed, 0, 0, false);
+    Pose2d pose = m_vision.getPose2d(m_currentPose, m_drive.getPose());
+    if(pose==null){
+      m_endCounter++;
+    }else{
+      m_endCounter = Math.max(0, m_endCounter-1);
+      m_currentPose = pose;
+      m_encoderPosition = getDist();
+      double dist1 = Math.abs(m_encoderPosition-m_encoderStart);
+      double dist2 = Math.sqrt(Math.pow(m_currentPose.getX()-m_startPose.getX(), 2) + Math.pow(m_currentPose.getY()-m_startPose.getY(), 2));
```
PandaNinja10: this line is too long

src/main/java/frc/robot/constants/ArmConstants.java
```diff
@@ -2,26 +2,23 @@
...
-  public static final double kShelfPosition = 0.4;
-  public static final double kIntakePosition = 0.3;
-  public static final double kMiddlePosition = 0.4;
-  public static final double kTopPosition = 0.5;
+  public static final double kinitialPosition = 0.0;
+  public static final double kshelfPosition = 0.4;
+  public static final double kintakePosition = 0.3;
+  public static final double kmiddlePosition = 0.4;
+  public static final double ktopPosition = 0.5;
+  public static final double klowPosition = kintakePosition;
```
PandaNinja10, line 23: The letter after k is suppose to be capitalized. also just there was no point to edit this file (on this branch) as it is all intake stuff not vision stuff.

src/main/java/frc/robot/constants/Constants.java
```diff
@@ -14,6 +14,7 @@ public final class Constants {
   public static final String kCanivoreCAN = "CANivore";
   public static final String kRioCAN = "rio";
 
+
```
PandaNinja10, line 17: random space in random file is not needed

src/main/java/frc/robot/constants/VisionConstants.java
```diff
@@ -0,0 +1,70 @@
...
+import edu.wpi.first.math.numbers.N3;
+import edu.wpi.first.math.util.Units;
+import frc.robot.Robot;
+import frc.robot.Robot.RobotId;
+
+
+
+public class VisionConstants {
+  public static ArrayList<Pair<String, Transform3d>> kCameras = new ArrayList<Pair<String, Transform3d>>(List.of(
+  ));
```
PandaNinja10, line 25: why is the "));" on a seprate line. the line of code is too long still but why at the very end, it is just confuseing.

src/main/java/frc/robot/constants/VisionConstants.java
```diff
@@ -0,0 +1,70 @@
...
+        )
+      ));
+    }
+  }
+
+  // How much to trust vision measurements normally
+  public static final Matrix<N3, N1> kBaseVisionPoseStdDevs = new MatBuilder<>(Nat.N3(), Nat.N1()).fill(
+    0.9, // x in meters (default=0.9)
+    0.9, // y in meters (default=0.9)
+    0.9 // heading in radians (default=0.9)
```
PandaNinja10, line 49: I don't think this is correct either the values or the comments are wrong

src/main/java/frc/robot/constants/VisionConstants.java
```diff
@@ -0,0 +1,70 @@
...
+    new AprilTag(2, new Pose3d(Units.inchesToMeters(610.77), Units.inchesToMeters(108.19), Units.inchesToMeters(18.22), new Rotation3d(0.0, 0.0, Math.PI))),
+    new AprilTag(3, new Pose3d(Units.inchesToMeters(610.77), Units.inchesToMeters(174.19), Units.inchesToMeters(18.22), new Rotation3d(0.0, 0.0, Math.PI))),
+    new AprilTag(4, new Pose3d(Units.inchesToMeters(636.96), Units.inchesToMeters(265.74), Units.inchesToMeters(27.38), new Rotation3d(0.0, 0.0, Math.PI))),
+    new AprilTag(5, new Pose3d(Units.inchesToMeters( 14.25), Units.inchesToMeters(265.74), Units.inchesToMeters(27.38), new Rotation3d(0.0, 0.0, 0.0))),
+    new AprilTag(6, new Pose3d(Units.inchesToMeters( 40.45), Units.inchesToMeters(174.19), Units.inchesToMeters(18.22), new Rotation3d(0.0, 0.0, 0.0))),
+    new AprilTag(7, new Pose3d(Units.inchesToMeters( 40.45), Units.inchesToMeters(108.19), Units.inchesToMeters(18.22), new Rotation3d(0.0, 0.0, 0.0))),
+    new AprilTag(8, new Pose3d(Units.inchesToMeters( 40.45), Units.inchesToMeters( 42.19), Units.inchesToMeters(18.22), new Rotation3d(0.0, 0.0, 0.0)))
```
PandaNinja10, line 68: all of these lines are too long maybe it is ok for constants but still it is still pretty hard to read.

Kyle-Eldridge: If I separated them, it might be even more confusing.

src/main/java/lib/controllers/GameController.java
```diff
@@ -14,6 +14,7 @@ public class GameController extends Controller {
       ALL_LEFT = get(DPad.LEFT).or(get(DPad.UP_LEFT)).or(get(DPad.DOWN_LEFT)),
       ALL_RIGHT = get(DPad.RIGHT).or(get(DPad.UP_RIGHT)).or(get(DPad.DOWN_RIGHT));
 
+
```
PandaNinja10, line 17: random space in random file not needed

src/main/java/frc/robot/subsystems/Module.java
```diff
@@ -18,13 +18,14 @@
...
-import frc.robot.constants.swerve.ModuleConstants;
-import frc.robot.constants.swerve.ModuleType;
 import frc.robot.util.LogManager;
 import frc.robot.util.MotorFactory;
 import lib.ctre_shims.TalonEncoder;
 
+import frc.robot.constants.swerve.ModuleConstants;
+import frc.robot.constants.swerve.ModuleType;
+import frc.robot.constants.swerve.DriveConstants;
+
```
PandaNinja10, line 28: moveing imports dosen't look nice and is unneeded

src/main/java/frc/robot/controls/Operator.java
```diff
@@ -1,25 +1,114 @@
...
+
+  private Drivetrain m_drive;
   private FourBarArm m_arm;
-  private Intake m_intake;
+  private Vision m_vision;
 
-  public Operator(FourBarArm arm, Intake intake){
-      m_arm = arm;
-      m_intake = intake;
+  public Operator(Drivetrain drive, FourBarArm arm, Vision vision){
```
PandaNinja10, line 43: intake is just gone now, please re-add the intake

src/main/java/frc/robot/subsystems/Drivetrain.java
```diff
@@ -107,7 +117,9 @@ public Drivetrain(ShuffleboardTab drivetrainTab, ShuffleboardTab swerveModulesTa
     };
     m_prevModule = m_modules[0];
     
-    m_odometry = new SwerveDriveOdometry(m_kinematics, m_pigeon.getRotation2d(), getModulePositions(), m_robotPose);
+    m_poseEstimator = new SwerveDrivePoseEstimator(m_kinematics, m_pigeon.getRotation2d(), getModulePositions(), new Pose2d());
```
PandaNinja10, line 120: this line is too long

src/main/java/frc/robot/subsystems/Drivetrain.java
```diff
@@ -107,7 +117,9 @@ public Drivetrain(ShuffleboardTab drivetrainTab, ShuffleboardTab swerveModulesTa
...
+    m_poseEstimator.setVisionMeasurementStdDevs(VisionConstants.kBaseVisionPoseStdDevs);
+
     m_rotationController.enableContinuousInput(-Math.PI, Math.PI);
     DoubleSupplier[] poseSupplier = {() -> getPose().getX(), () -> getPose().getY(), () -> getPose().getRotation().getRadians()};
```
PandaNinja10, line 124: not your fault but this line is also too long

PandaNinja10:
Remove Operator Slection stuff as it dosen't work and is just cultter.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -152,6 +164,14 @@ public void addTestCommands() {
     m_testTab.add("Heading PID", new TestHeadingPID(m_drive, testEntry));
     m_testTab.add("Steer angle", new TestSteerAngle(m_drive, testEntry));
     m_testTab.add("Odometry Test", new OdometryTestCommand(m_drive, new Transform2d(new Translation2d(1,1), new Rotation2d(Math.PI))));
+    m_testTab.add("Test vision (forward)", new TestVision(0.2, m_drive, m_vision));
+    m_testTab.add("Test vision (backward)", new TestVision(-0.2, m_drive, m_vision));
+    m_testTab.add("Test vision 2 (forward)", new VisionReturnTest(0.2, 3, m_drive, m_vision));
+    m_testTab.add("Test vision 2 (backward)", new VisionReturnTest(-+ 0.2, 3, m_drive, m_vision));
```
PandaNinja10, line 170: update titles names

src/main/java/frc/robot/commands/AlignToColumn.java
```diff
@@ -0,0 +1,29 @@
...
+
+  private final Drivetrain m_drive;
+  
+  public AlignToColumn(Drivetrain drive) {
+    m_drive = drive;
+    addRequirements(drive);
+    addCommands(
+      new PathPlannerCommand(new ArrayList<PathPoint>(List.of(
+        PathPoint.fromCurrentHolonomicState(m_drive.getPose(), m_drive.getChassisSpeeds()),
+        new PathPoint(Operator.selectedNode.scorePose.getTranslation(), Operator.selectedNode.scorePose.getRotation())
```
PandaNinja10, line 25: this is reliant on Operator please make it so it is not relient on operator as we are removeing for now.

src/main/java/frc/robot/commands/DefaultDriveCommand.java
```diff
@@ -1,5 +1,7 @@
 package frc.robot.commands;
 
+import edu.wpi.first.wpilibj.DriverStation;
+import edu.wpi.first.wpilibj.DriverStation.Alliance;
```
PandaNinja10, line 4: no need for random imports

src/main/java/frc/robot/commands/MoveToSelectedPose.java
```diff
@@ -0,0 +1,35 @@
...
+    double x = Operator.selectedNode.scorePose.getX();
+    double y = Operator.selectedNode.scorePose.getY();
+    double rot = Operator.selectedNode.scorePose.getRotation().getRadians();
+
+    m_drive.runChassisPID(x, y, rot);
+  }
+  
+  @Override
+  public void end(boolean interrupted) {
+    m_drive.drive(0.0, 0.0, 0.0, false);
```
PandaNinja10, line 33: there is a .stop() method might want to use that

src/main/java/frc/robot/commands/AlignToColumn.java
Kyle-Eldridge: This command doesn't work without Operator. Also, it doesn't cause an error because I didn't comment out the selectedNode in Operator.

src/main/java/frc/robot/util/Node.java
```diff
@@ -0,0 +1,80 @@
...
+            x = tag.getX()+Units.inchesToMeters(14.25/2)*(alliance==Alliance.Red?-1:1);
+            z = 0;
+            break;
+            case(2):
+            x = tag.getX()-Units.inchesToMeters((31.625-14.25)/2)*(alliance==Alliance.Red?-1:1);
+            z = Units.inchesToMeters(type==Types.CUBE?23.5:34);
+            break;
+            case(3):
+            x = tag.getX()-Units.inchesToMeters(39.75-14.25)*(alliance==Alliance.Red?-1:1);
+            z = Units.inchesToMeters(type==Types.CUBE?35.5:46);
```
PandaNinja10, line 68: it would be nice to explain a bit of what the magic numbers are doing

src/main/java/frc/robot/util/Node.java
```diff
@@ -0,0 +1,80 @@
...
+            break;
+            default:
+            throw(new IllegalArgumentException("Row had to be 1, 2, or 3"));
+        }
+        // Y coordinate in meters
+        double y = tag.getY()+(column%3==2?0:Units.inchesToMeters(22))*((column%3==0)^(alliance==Alliance.Red)?-1:1);
+        // The pose of the node
+        pose=new Pose3d(x, y, z, tag.getRotation());
+        // The pose the robot goes to to score
+        scorePose=new Pose2d(tag.getX()+Units.inchesToMeters(14.25+VisionConstants.kGridDistance)*(alliance==Alliance.Red?-1:1), y, new Rotation2d(alliance==Alliance.Red?0:Math.PI));
```
PandaNinja10, line 78: it would be nice to explain a bit of what the magic numbers are doing

PandaNinja10, line 78: spaces between equal signs

src/main/java/frc/robot/commands/vision/TestVision.java
PandaNinja10: you should have a way to temperaraly disable vison inputs to odometry for the test

src/main/java/frc/robot/commands/vision/VisionReturnTest.java
```diff
@@ -0,0 +1,122 @@
...
+    m_speed=speed;
+    m_distanceToMove=distance;
+  }
+
+  private double getDist(){
+    return 
+      m_drive.m_modules[0].getDrivePosition()/4+
+      m_drive.m_modules[1].getDrivePosition()/4+
+      m_drive.m_modules[2].getDrivePosition()/4+
+      m_drive.m_modules[3].getDrivePosition()/4;
```
PandaNinja10, line 44: Please use odometry for the distance rather than the individual encoder positions

src/main/java/frc/robot/commands/vision/VisionReturnTest.java
```diff
@@ -0,0 +1,122 @@
...
+        }else if(dist3<m_distanceToMove && m_direction > 0){
+          m_direction *= -0.8;
+          m_turns++;
+        }
+        if(m_turns>10){
+          m_endCounter+=2;
+        }
+      }
+    }
+  }
```
PandaNinja10, line 100: some comments would be nice

src/main/java/frc/robot/constants/ArmConstants.java
```diff
@@ -24,4 +20,5 @@ public class ArmConstants {
   public static final double kIntakePosition = 0.3;
   public static final double kMiddlePosition = 0.4;
   public static final double kTopPosition = 0.5;
+  public static final double kLowPosition = kIntakePosition;
```
PandaNinja10, line 23: just don't change this file not the place to do this.

src/main/java/frc/robot/commands/MoveToSelectedPose.java
```diff
@@ -0,0 +1,34 @@
...
+    double rot = Operator.selectedNode.scorePose.getRotation().getRadians();
+
+    m_drive.runChassisPID(x, y, rot);
+  }
+  
+  @Override
+  public void end(boolean interrupted) {
+    m_drive.stop();
+  }
+}
```
PandaNinja10, line 34: Richie want this PR to be purely Odometry and geting pose no driveing for alignment, operator config, etc.

src/main/java/frc/robot/controls/Operator.java
```diff
@@ -1,25 +1,119 @@
...
+        selectValues[0]=pressValue;
+      } else if (selectValues[1]==0) {
+        selectValues[1]=pressValue;
+      } else {
+        selectValues[2] = pressValue;
+        selectTime = 1;
+
+        selectValue(0, selectValues[0]);
+      }
+    }
```
PandaNinja10, line 116: Operator config might be better on a seperate PR

src/main/java/frc/robot/commands/vision/Align.java
```diff
@@ -0,0 +1,95 @@
...
+    return p2.getRotation().getRadians();
+  }
+
+  /**
+   * Initializes the command
+   */
+  @Override
+  public void initialize(){
+    m_angle=m_setpoint-2*Math.PI;
+    m_angle=getAngle();
```
PandaNinja10, line 53: why is m_angle set twice. also spaces please

src/main/java/frc/robot/commands/vision/Align.java
```diff
@@ -0,0 +1,95 @@
...
+
+  private double getAngle(){
+    ArrayList<EstimatedRobotPose> p = m_vision.getEstimatedPoses(m_drive.getPose());
+    Pose2d p2;
+    if(p.size()==0){
+      return m_angle;
+    }else if(p.size()==1){
+      p2=p.get(0).estimatedPose.toPose2d();
+    }else{
+      p2 = new Pose2d(p.get(0).estimatedPose.getX()/2+p.get(1).estimatedPose.getX()/2, p.get(0).estimatedPose.getY()/2+p.get(1).estimatedPose.getY()/2, new Rotation2d(p.get(0).estimatedPose.toPose2d().getRotation().getRadians()/2+p.get(1).estimatedPose.toPose2d().getRotation().getRadians()/2));
```
PandaNinja10, line 42: this line is too long

PandaNinja10, line 42: getRadians will output a value between -Pi and Pi so the avrageing radians might have a few probolems. For example the first pose has just under Pi andthe second is jut over -Pi. The avrage is around 0 but the acutal value is aproxamitly Pi or Pi.

src/main/java/frc/robot/commands/vision/Align.java
```diff
@@ -0,0 +1,95 @@
...
+import org.photonvision.EstimatedRobotPose;
+
+import edu.wpi.first.math.geometry.Pose2d;
+import edu.wpi.first.math.geometry.Rotation2d;
+import edu.wpi.first.math.util.Units;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.subsystems.Drivetrain;
+import frc.robot.util.Vision;
+
+public class Align extends CommandBase{
```
PandaNinja10, line 15: I think align is pretty vauge and a better name such as "apriltagAngleAlignTest" would be more discriptive

src/main/java/frc/robot/util/Node.java
```diff
@@ -0,0 +1,80 @@
...
+public class Node {
+    // Possible node types
+    public enum Types {CONE, CUBE, HYBRID};
+
+    public final Alliance alliance;
+    public final int row;
+    public final int column;
+    public final Pose3d pose;
+    public final Pose2d scorePose;
+    public final Types type;
```
PandaNinja10: add "m_" to these variables

src/main/java/frc/robot/util/Node.java
```diff
@@ -0,0 +1,80 @@
...
+     *  Grid column from left to right (1-9)
+     */
+    public Node(Vision vision, Alliance alliance, int row, int column){
+        this.alliance=alliance;
+        this.row=row;
+        this.column=column;
+        type = row==1?Types.HYBRID:column%3==2?Types.CUBE:Types.CONE;
+
+        // Which April tag this node is closest to
+        Pose3d tag = vision.getTagPose((column-1)/3+(alliance==Alliance.Red?1:6));
```
PandaNinja10: I don't think holding the node positions relitive to the tag is the best way to do it. Additional  maybe this file and the align command should be in a seprate PR

src/main/java/frc/robot/commands/AlignToColumn.java
PandaNinja10: you can make it non-dependent on operator by adding a node paramater

src/main/java/frc/robot/commands/AlignToColumn.java
```diff
@@ -0,0 +1,30 @@
...
+public class AlignToColumn extends SequentialCommandGroup {
+
+  private final Drivetrain m_drive;
+  
+  public AlignToColumn(Drivetrain drive) {
+    m_drive = drive;
+    addRequirements(drive);
+    addCommands(
+      new PathPlannerCommand(new ArrayList<PathPoint>(List.of(
+        PathPoint.fromCurrentHolonomicState(m_drive.getPose(), m_drive.getChassisSpeeds()),
```
PandaNinja10, line 25: The way this is writen it only gets the starting position on construction. I would advise against this as this means that the align to coulum command needs to be made right as the it is being called to run. Addtional you can prorobly use the GoToPoseCommand in Auto-Paths-and-Path-Planner branch later.

src/main/java/frc/robot/commands/AlignToColumn.java
```diff
@@ -0,0 +1,30 @@
...
+        new PathPoint(new Node().scorePose.getTranslation(), new Node().scorePose.getRotation())
+      )), m_drive)
+    );
+  }
+}
```
PandaNinja10, line 30: could be in another PR

src/main/java/frc/robot/util/Node.java
```diff
@@ -0,0 +1,80 @@
...
+        throw(new IllegalArgumentException("Row had to be 1, 2, or 3"));
+    }
+    // Y coordinate in meters
+    double y = tag.getY()+(column%3==2?0:Units.inchesToMeters(22))*((column%3==0)^(alliance==Alliance.Red)?-1:1);
+    // The pose of the node
+    pose = new Pose3d(x, y, z, tag.getRotation());
+    // The pose the robot goes to to score
+    scorePose = new Pose2d(tag.getX()+Units.inchesToMeters(14.25+VisionConstants.kGridDistance)*(alliance==Alliance.Red?-1:1), y, new Rotation2d(alliance==Alliance.Red?0:Math.PI));
+  }
+}
```
PandaNinja10, line 80: Could be moved to a seprate PR for now

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -152,6 +164,14 @@ public void addTestCommands() {
     m_testTab.add("Heading PID", new TestHeadingPID(m_drive, testEntry));
     m_testTab.add("Steer angle", new TestSteerAngle(m_drive, testEntry));
     m_testTab.add("Odometry Test", new OdometryTestCommand(m_drive, new Transform2d(new Translation2d(1,1), new Rotation2d(Math.PI))));
+    m_testTab.add("Test vision (forward)", new TestVision(0.2, m_drive, m_vision));
+    m_testTab.add("Test vision (backward)", new TestVision(-0.2, m_drive, m_vision));
+    m_testTab.add("Vision return test (forward)", new VisionReturnTest(0.2, 3, m_drive, m_vision));
+    m_testTab.add("Vision return test (backward)", new VisionReturnTest(-+ 0.2, 3, m_drive, m_vision));
```
PandaNinja10, line 170: pretty "-+ 0.2" a typo becuse it is pretty confuseing, just put "-0.2"

src/main/java/frc/robot/constants/VisionConstants.java
PandaNinja10: the variable has been removed

src/main/java/frc/robot/commands/PathPlannerCommand.java
PandaNinja10: Removed to be reimplemented in a later PR

src/main/java/frc/robot/commands/vision/AngleAlignTest.java
```diff
@@ -0,0 +1,104 @@
+package frc.robot.commands.vision;
```
PandaNinja10, line 1: I noticed you pretty frequently use really one or two character variable wich are unclear please rename them to something more clear so it is easier to understand

src/main/java/frc/robot/commands/vision/AngleAlignTest.java
```diff
@@ -0,0 +1,104 @@
...
+  public void execute(){
+    m_angle = getAngle();
+    double a = m_angle;
+    if(a<0&&m_setpoint>0){
+      a += 2*Math.PI;
+    }else if(a>0&&m_setpoint<0){
+      a -= 2*Math.PI;
+    }
+    double speed = a>m_setpoint?-0.1:0.1;
+    m_drive.drive(0, 0, -speed, false);
```
PandaNinja10, line 79: this seams a bit janky and hard to fallow. I suggust you use the rotational PID stored in drivetrain

src/main/java/frc/robot/commands/vision/AngleAlignTest.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+    return p2.getRotation().getRadians();
+  }
+
+  /**
+   * Initializes the command
+   */
+  @Override
+  public void initialize(){
+    m_angle = m_setpoint-2*Math.PI;
```
PandaNinja10, line 62: why is this in initializaiton rather than the constuctor

src/main/java/frc/robot/commands/vision/AngleAlignTest.java
```diff
@@ -0,0 +1,104 @@
...
+        p.get(0).estimatedPose.getX() / 2 + p.get(1).estimatedPose.getX() / 2, 
+        p.get(0).estimatedPose.getY() / 2 + p.get(1).estimatedPose.getY() / 2, 
+        new Rotation2d( Functions.modulusMidpoint(
+          p.get(0).estimatedPose.toPose2d().getRotation().getRadians(),
+          p.get(1).estimatedPose.toPose2d().getRotation().getRadians(),
+          -Math.PI, Math.PI
+        ))
+      );
+    }
+    return p2.getRotation().getRadians();
```
PandaNinja10, line 54: why do you store the pose rather than returning the radian directly. especial for when you get the rotation anyway in the  avrage angle anyway.

src/main/java/frc/robot/subsystems/Drivetrain.java
```diff
@@ -205,20 +247,54 @@ public void setChassisSpeeds(ChassisSpeeds chassisSpeeds) {
    * @param rot the angle to move to, in radians
    */
   public void runChassisPID(double x, double y, double rot) {
-    double xSpeed = m_xController.calculate(m_odometry.getPoseMeters().getX(), x);
-    double ySpeed = m_yController.calculate(m_odometry.getPoseMeters().getY(), y);
+    double xSpeed = m_xController.calculate(m_poseEstimator.getEstimatedPosition().getTranslation().getX(), x);
+    double ySpeed = m_yController.calculate(m_poseEstimator.getEstimatedPosition().getTranslation().getY(), y);
```
PandaNinja10, line 251: you don't need to get the translaion, you can get the X and Y directly from the pose.

src/main/java/frc/robot/commands/vision/TestVision.java
```diff
@@ -0,0 +1,92 @@
...
+import frc.robot.subsystems.Drivetrain;
+import frc.robot.util.Vision;
+
+public class TestVision extends CommandBase{
+  private Drivetrain m_drive;
+  private Vision m_vision;
+  private Translation2d m_visionStartTranslation, m_driveStartTranslation;
+  private Pose2d m_currentPose = null;
+  private int m_endCounter = 0;
+  private int m_printCounter=0;
```
PandaNinja10, line 15: I think it might be a good idea to use a timer rather than counting cycles. It would proobly be easier to understand and easier to adjust. Instead of printing out your result every few cycles it might be a good idea to use shuffleboard.

src/main/java/frc/robot/commands/vision/TestVision.java
```diff
@@ -0,0 +1,92 @@
...
+  private double m_speed;
+
+  //How many frames it has to not see anything to end the command
+  private static final int endDelay = 5;
```
PandaNinja10, line 19: inaccurate comment, this is how many frames it has not seen any thing but how many rio cycles it has acompished. If the rio runtime and april tag update frequencies were the same it wouldn't be a problem but becuse they are diffrent this is problematic.

PandaNinja10:
It is really confuseing to have the unit test file called "VisonTest" and the runnable vison test called "TestVison"

src/main/java/frc/robot/commands/vision/TestVisionAlignment.java
```diff
@@ -0,0 +1,61 @@
...
+  @Override
+  public void end(boolean interrupted) {
+    //print vision angle
+    System.out.printf("\nExact angle: %.4f degrees\n", Units.radiansToDegrees(getAngle()));
+    m_drive.stop();
+  }
+
+  @Override
+  public boolean isFinished(){
+    return Math.abs(MathUtil.angleModulus(getAngle() - m_setpoint)) < TestConstants.kHeadingError;
```
PandaNinja10, line 52: you could use the heading PID is at setpoint method

src/main/java/frc/robot/commands/vision/TestVisionAlignment.java
```diff
@@ -0,0 +1,61 @@
...
+  /**
+   * Turns the robot
+   * PID broke the classbot, so this doesn't use it
+   */
+  @Override
+  public void execute() {
+    m_mostRecentAngle = getAngle();
+    double speed = m_drive.getRotationController().calculate(m_mostRecentAngle, m_setpoint);
+    m_drive.drive(0, 0, speed, false);
+  }
```
PandaNinja10, line 41: remember to reset the PID before useing it

src/main/java/frc/robot/commands/vision/TestVisionDistance.java
```diff
@@ -0,0 +1,88 @@
...
+  private Pose2d m_currentPose = null;
+
+  private double m_speed;
+
+  private final Timer m_endTimer = new Timer();
+  private final Timer m_printTimer = new Timer();
+
+  // TODO: Should these values be in a constant file?
+  // How many seconds of not seeing april tag before ending the command
+  private static final int kEndDelay = 5;
```
PandaNinja10, line 26: it it need so not see a aprial tag for 5 seconds to stop, I don't think so

PandaNinja10 (approved):
So after cleaning and removeing features I think we are a point where it is clean enough that we can merge it. Only major things that need to be added are the positions of the comp cameras. transofmations and any patches later down the line. Over all this is good enough to merge.

PR #18, opened 2023-02-13
PowerPlay April Tag locations

GLRoylance:
## Overview

This PR adds the AprilTag locations to the constants directory.

## Tests Ran

Compiles. SimGUI runs.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
 plugins {
     id "java"
-    id "edu.wpi.first.GradleRIO" version "2023.1.1"
+    id "edu.wpi.first.GradleRIO" version "2023.2.1"
     id "com.peterabeles.gversion" version "1.10"
 }
 
--- a/src/main/java/frc/robot/Robot.java
+++ b/src/main/java/frc/robot/Robot.java
@@ -85,13 +85,17 @@ public class Robot extends TimedRobot {
 
   /** This function is called periodically when the robot is disabled */
   @Override
-  public void disabledPeriodic() {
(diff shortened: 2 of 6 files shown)


AnthonyFurman:
is there no file for loadFromResource() for this?

Also, if they are the same for powerplay and for charged up, is there a reason to call it power play? I feel like charged up would be less confusing.

I'm also not sure that the unit tests are useful.

GLRoylance:
I did not find a loadFromResource file. I expected kDefault to work.

My brain bubble. It should be ChargedUp.

If the locations are loaded from a file, it will make sure the data is
accurate. It will assert if loadFromResource uses kRapidReact or kDefault.
There may be a subsequent release.

Many unit tests may look stupid, but stupid tests are fast.

Jerry

GLRoylance:
Another way to look at seemingly silly unit tests. If you were developing some code to use the April Tag Field Layout, wouldn't you try accessing one of the tags and making sure you got the expected location? If that silly test didn't work, it would be pointless to use the field layout.

AnthonyFurman:
> wouldn't you try accessing one of the tags and making sure you got the expected location
Yes, I can look at the file to do that. 

Anyway seems like there is a file, so I think we agree that is preferred. I think to update this will work: Ctrl-Shift-P -> Check for wpi updates

GLRoylance:
edited build.gradle to use 2023.2.1.

The kDefault and k2023ChargedUp values now validate.

AnthonyFurman:
I've updated the main branch to have the latest version.

PR #21, opened 2023-02-14
Four bar problem fix

GLRoylance:
## Overview

This PR deletes an unused variable (kills 1 VS reported problem). It adds JavaDoc comments to several methods. It adds explanation for RelativeEncoder scaling. It sets some temporary constant values (marked with TODO:).

## Design Doc

Relatively minor changes.

## Tests Ran

SimGUI

### DIFF
--- a/src/main/java/frc/robot/constants/ArmConstants.java
+++ b/src/main/java/frc/robot/constants/ArmConstants.java
@@ -1,21 +1,21 @@
 package frc.robot.constants;
 
 public class ArmConstants {
-  // arm ids - TBD
-  public static final int motorID = 1;
+  // TODO: arm ids - TBD (fake using 5 -- 1-4 have been taken)
+  public static final int motorID = 5;
 
-  // pid values - TBD
+  // TODO: PID values - TBD (fake values for now)
   public static final double kP = -1;
-  public static final double kI = -1;
-  public static final double kD = -1;
-  public static final double kTolerance = -1;
-  public static final double minMotorPower = -1;
-  public static final double maxMotorPower = -1;
+  public static final double kI = 0.0;
+  public static final double kD = 0.0;
+  public static final double kTolerance = 0.02;
+  public static final double minMotorPower = -1.0;
+  public static final double maxMotorPower = +1.0;
 
-  // distance values - TBD
-  public static final double initialPosition = 0;
-  public static final double shelfPosition = -1;
-  public static final double intakePosition = -1;
-  public static final double middlePosiiton = -1;
-  public static final double topPosition = -1;
+  // TODO: distance values - TBD (fake values for now)
+  public static final double initialPosition = 0.0;
+  public static final double shelfPosition = 0.4;
+  public static final double intakePosition = 0.3;
+  public static final double middlePosiiton = 0.4;
+  public static final double topPosition = 0.5;
 }
--- a/src/main/java/frc/robot/constants/ArmConstants.java
+++ b/src/main/java/frc/robot/constants/ArmConstants.java
@@ -5,7 +5,7 @@ public class ArmConstants {
   public static final int motorID = 5;
 
   // TODO: PID values - TBD (fake values for now)
-  public static final double kP = -1;
+  public static final double kP = 1.0;
   public static final double kI = 0.0;
   public static final double kD = 0.0;
   public static final double kTolerance = 0.02;
--- a/src/main/java/frc/robot/constants/ArmConstants.java
+++ b/src/main/java/frc/robot/constants/ArmConstants.java
@@ -1,5 +1,7 @@
 package frc.robot.constants;
 
+import edu.wpi.first.math.util.Units;
+
 public class ArmConstants {
   // TODO: arm ids - TBD (fake using 5 -- 1-4 have been taken)
   public static final int motorID = 5;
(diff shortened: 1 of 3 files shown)


src/main/java/frc/robot/constants/ArmConstants.java
```diff
@@ -1,21 +1,21 @@
...
 
-  // pid values - TBD
-  public static final double kP = -1;
-  public static final double kI = -1;
-  public static final double kD = -1;
-  public static final double kTolerance = -1;
-  public static final double minMotorPower = -1;
-  public static final double maxMotorPower = -1;
+  // TODO: PID values - TBD (fake values for now)
+  public static final double kP = 1.0;
```
AnthonyFurman, line 8: I think using a P value of zero, starting out, would be safer.

src/main/java/frc/robot/subsystems/FourBarArm.java
```diff
@@ -14,32 +14,57 @@ public class FourBarArm extends SubsystemBase {
...
   @Override
   public void periodic() {
-    m_motor.set(MathUtil.clamp(m_pid.calculate(m_encoder.getPosition()), ArmConstants.minMotorPower, ArmConstants.maxMotorPower));
+    // calculate the PID power level
+    double pidPower = m_pid.calculate(m_encoder.getPosition());
+    // calculate the feedforward power (nothing for now)
+    double feedforwardPower = 0.0;
+
+    // set the motor power
+    m_motor.set(MathUtil.clamp(pidPower + feedforwardPower, ArmConstants.minMotorPower, ArmConstants.maxMotorPower));
```
AnthonyFurman, line 61: I don't think this is necessary. Yichen is working on the four bar arm feedforward. Please talk to us if you want to suggest changes like this.

src/main/java/frc/robot/constants/ArmConstants.java
GLRoylance: With PID = (0, 0, 0), a simulation would never converge.

src/main/java/frc/robot/constants/ArmConstants.java
```diff
@@ -1,21 +1,21 @@
...
+  public static final double kI = 0.0;
+  public static final double kD = 0.0;
+  public static final double kTolerance = 0.02;
+  public static final double minMotorPower = -1.0;
+  public static final double maxMotorPower = +1.0;
```
AnthonyFurman, line 13: for safety I would use lower values

src/main/java/frc/robot/controls/Operator.java
```diff
@@ -15,7 +15,7 @@ public class Operator {
    */
   public static void configureControls(FourBarArm arm) {
     operator.get(Button.Y).onTrue(new ExtendToPosition(arm, ArmConstants.topPosition));
-    operator.get(Button.X).onTrue(new ExtendToPosition(arm, ArmConstants.middlePosiiton));
+    operator.get(Button.X).onTrue(new ExtendToPosition(arm, ArmConstants.middlePosition));
     operator.get(Button.A).onTrue(new ExtendToPosition(arm, ArmConstants.intakePosition));
     operator.get(Button.Y).onTrue(new ExtendToPosition(arm, ArmConstants.shelfPosition));
```
AnthonyFurman, line 20: @MagicEmu I know you were made aware of the Y button issue

src/main/java/frc/robot/constants/ArmConstants.java
AnthonyFurman: Yes, simulation is working on a different branch.

AnthonyFurman:
I would prefer @MagicEmu make these changes.
In the future, @GLRoylance , this feedback would be useful on the PR.

src/main/java/frc/robot/subsystems/FourBarArm.java
GLRoylance: Bare code is often difficult to understand. I've added comments that explain the research I did to understand the code.

For PIDF, WPILIB has bottled up the PID but not the F. Separate them to make the pieces clear.

src/main/java/frc/robot/constants/ArmConstants.java
GLRoylance: Safety might be better served by limiting the power to a small value such as 0.05.

I've been looking at the simulation branch. I like that it is using both the physics and mech2d. I also like the close method.

src/main/java/frc/robot/constants/ArmConstants.java
GLRoylance: min/max power reduced to 0.05.

GLRoylance: kTolerance of 0.02 radian is about 1 degree.

src/main/java/frc/robot/subsystems/FourBarArm.java
AnthonyFurman: Yes, I don't think the comments are really necessary to explain that research. We have the WPI documentation, and I did some work to explain PID + FF to the students during last semester.

src/main/java/frc/robot/constants/ArmConstants.java
AnthonyFurman: kTolerance is fine, mainly meant min/max reduced. Though that being said, to be more clear, you can do Units.degreesToRadians(1) which is easier to understand and emphasizes the units used.

AnthonyFurman (approved):

PR #25, opened 2023-02-19
Constant switching fix

omegaladon:
## Overview
Added fix for robotID key and changed constant switching to be a method called in robotcontainer
UNTESTED CODE (due to this being done at home, someone please test before merging)


PandaNinja10 (approved):
My bad, I didn't catch this on the last PR

PR #30, opened 2023-02-23
Mechanism

GLRoylance:
## Overview

This PR provides Mechanism2d display of elevator, four bar, and intake.

## Tests Ran

SimGUI


AnthonyFurman:
Could the name be changed to MechanismSimDisplay or something similar? "Mechanism" is very vague.

In the future, I would prefer a student do this kind of programming, even if you are just telling them what to do. There are free programmers.
Maybe we can discuss some way to get more communication about the changes you want. My job is managing the programming students, so I give them tasks about what I think needs to be done. I think it would be helpful to also know what tasks you would want done, so I can have students work on them. Otherwise, you and a student might end up working on the same thing (which has already happened to a certain extent), and it would be more efficient. Might be after HPR but emails or slack would work.

src/main/java/frc/robot/Robot.java
```diff
@@ -89,6 +89,7 @@ public void robotPeriodic() {
   @Override
   public void disabledInit() {
     CommandScheduler.getInstance().cancelAll();
+    // TODO: This method was called in robotInit(); is it needed here?
```
PandaNinja10, line 92: yes it is needed there to create new pathplanner commands as they only run once per object

src/main/java/frc/robot/Robot.java
```diff
@@ -89,6 +89,7 @@ public void robotPeriodic() {
...
     m_robotContainer.autoChooserUpdate();
```
PandaNinja10, line 93: yes it is needed there to create new pathplanner commands as they only run once per object

AnthonyFurman:
I started making some changes for the wrist but realized there might need to be a bigger scope of changes. The values need to be checked for the wrist, I just put in ones so that it wouldn't look like it was colliding. I also added coloring the intake based on intake mode in lieu of showing spinning. 
Additionally, setDistanceToGrid doesn't appear to work. The elevator extends to some very high position when the grid distance is set.

src/main/java/frc/robot/util/DrawMechanism.java
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    private void addCommands(DrawMechanism mechanism) {
+        // some commands to move the Mech2d diagram
+        SmartDashboard.putData("wrist in", new InstantCommand(() -> mechanism.setWristAngle(WristConstants.kStowPos)));
+        SmartDashboard.putData("wrist out", new InstantCommand(() -> mechanism.setWristAngle(WristConstants.kIntakeCubePos)));
+        SmartDashboard.putData("elevator down", new InstantCommand(() -> mechanism.setElevatorExtension(0.1)));
+        SmartDashboard.putData("elevator up", new InstantCommand(() -> mechanism.setElevatorExtension(1.3)));
+        SmartDashboard.putData("drivetrain in", new InstantCommand(() -> mechanism.setDistanceToGrid(16.5)));
+        SmartDashboard.putData("drivetrain away", new InstantCommand(() -> mechanism.setElevatorExtension(48)));
```
GLRoylance, line 200: "drivetrain away" sets elevator extension to 48 meters

AnthonyFurman: right, no more late night coding...

AnthonyFurman (approved):

PR #35, opened 2023-02-25
fix offset constants

PandaNinja10:
I some how didn't copy and paste the constants corectly to fix swerve offset. Riche how did you not catch it, the variable names were wrong?


AnthonyFurman (approved):
smh blame everything on richie

richtan (approved):

PR #38, opened 2023-02-26
Pose commands

PandaNinja10:
## Overview

This PR is moveing some features from the Auto-Paths-and-Path-Planner branch so vison can have commands to work with

## Design Doc

## Tests Ran

Go To Pose, Transform Pose, and thier Test CounterParts have been run and operated as expected. The odometry drift is roughly 0.4m after a decent amont of teleop driveing. the translation PID dosen't have a notable error from it's esitmated position.


src/main/java/frc/robot/commands/GoToPose.java
```diff
@@ -0,0 +1,67 @@
...
+  private Supplier<Pose2d> m_poseSupplier;
+  private double m_startTime;
+  private Pose2d m_finalPose;
+  private Pose2d m_error;
+  
+  public GoToPose(Drivetrain drive, Pose2d pose) {
+    m_drive = drive; 
+    // finalPose is position after robot moves from current position-- startPose-- by the values that are inputted-- distanceToMove
+    m_finalPose = pose;
+    addRequirements(drive);
```
AnthonyFurman, line 28: Could just do m_poseSupplier = () -> pose
That would make the pose supplier a "supplier" but it always supplies the same pose. Just cleaner so variables aren't null

AnthonyFurman: Also the comment is unclear

src/main/java/frc/robot/commands/GoToPose.java
```diff
@@ -0,0 +1,67 @@
...
+  }
+  
+  @Override
+  public void execute() {
+    m_drive.runChassisPID(m_finalPose.getX(), m_finalPose.getY(), m_finalPose.getRotation().getRadians()); 
+  }
+
+  @Override
+  public boolean isFinished() {
+    // TODO: the current PID values don't allow the command to finish
```
AnthonyFurman, line 50: In you PR comment you say there is no notable error on the chassis PID, so is this comment still accurate?

src/main/java/frc/robot/commands/GoToPose.java
```diff
@@ -0,0 +1,67 @@
...
+import edu.wpi.first.wpilibj.Timer;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.constants.TestConstants;
+import frc.robot.subsystems.Drivetrain;
+
+/**
+ * Tests the odometry of the robot by driving a certain distance and calculating the error.
+ */
+public class GoToPose extends CommandBase {
+
```
AnthonyFurman, line 17: This command seems to be the same as the odometery test command. So i would either remove the odometery test command or this one.
The only difference in the commands is that one takes a Transform2d, a distance to move relative to the robot, and this one takes an absolute field pose. They can be made the same command by just adding another constructor that takes a robot relative Transform2d

src/main/java/frc/robot/commands/PoseTransform.java
```diff
@@ -0,0 +1,72 @@
...
+import edu.wpi.first.wpilibj.Timer;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.constants.TestConstants;
+import frc.robot.subsystems.Drivetrain;
+
+/**
+ * Tests the odometry of the robot by driving a certain distance and calculating the error.
+ */
+public class PoseTransform extends CommandBase {
+
```
AnthonyFurman, line 17: See my comment on the other command. This seems to be the same command as odometerTest. Just make one command with different constructors.

AnthonyFurman: Ok I looked at it now and it looks like one is a test and the other is a actual command. I would just make the test call the actual one. And, you only need 1 test command, one actual command

src/main/java/frc/robot/commands/GoToPose.java
```diff
@@ -0,0 +1,104 @@
...
+  private boolean m_doNotEnd = false;
+  
+  /**
+   * Returns a command that goes to a position based on shuffleboard inputs
+   * DO NOT USE IF SHUFFLEBOARD IS DISABLED 
+   * @param drive drivetrain to be used for the command
+   * @param relativeToRobot should pose be converted to a transformation
+   * @param doNotEnd should the robot not end for tunning purposes
+   */
+  public GoToPose(Drivetrain drive, boolean relativeToRobot, boolean doNotEnd){
```
AnthonyFurman, line 34: You can use PerpetualCommand(commandToRun) which will run the command but ignore the isFinished, instead of using the extra boolean param. Or, you can add the .perpetually() decorator to do the same thing

AnthonyFurman: Basically that would remove the doesNotEnd parameter

src/main/java/frc/robot/commands/GoToPose.java
```diff
@@ -0,0 +1,104 @@
...
+  public void execute() {
+    m_drive.runChassisPID(m_finalPose.getX(), m_finalPose.getY(), m_finalPose.getRotation().getRadians()); 
+  }
+
+  @Override
+  public boolean isFinished() {
+    if (m_doNotEnd) return false;
+    double errorMarginMeters = TestConstants.kTranslationError;
+    double errorMarginRadians = TestConstants.kHeadingError;
+    m_error = m_drive.getPose().relativeTo(m_finalPose);
```
richtan, line 88: use Translation.getDistance()

src/main/java/frc/robot/commands/GoToPose.java
PandaNinja10: Perpertual Command is deprecated and will be removed in the future so I don't think we should use it.

src/main/java/frc/robot/commands/GoToPose.java
PandaNinja10: This spcific PR is still to be tested and the PID values should be valided before mergeing as they have been tuned more on Auto-Paths-and-Path-Planner

PandaNinja10:
These changes will be in the auto PR

AnthonyFurman (dismissed):

PR #40, opened 2023-02-26
auto command refresh

PandaNinja10:
refreshes the auto choser commands by createing new commands everytime you get the auto command.


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -147,6 +147,7 @@ public void initDriveYaw(boolean force) {
    * @return the command to run in autonomous
    */
   public Command getAutonomousCommand() {
+    autoChooserUpdate();
     return m_autoCommand.getSelected();
```
AnthonyFurman, line 151: How much stuff does this add to shuffleboard? Last year when we did this it took so long to add all the stuff, the robot should wait for a few seconds after starting auto

richtan (approved):

src/main/java/frc/robot/RobotContainer.java
PandaNinja10: It only replaces the values stored in the auto chooser so right now it is only the do nothing but it will increse as we add routines. we could prorobly add this to the end of auto so it dosen't delay the robot starting if it is a problem

PR #41, opened 2023-02-27
Charge Station Vision PR

Kyle-Eldridge:
## Overview

This PR will make the robot trust vision more during auto if it goes over the charge station because the encoders will think it goes farther than it actually does.

## Design Doc

None. This is a small PR.

## Tests Ran

I don't have the robot for testing, so I can't run any tests. I only changed a few things, and they won't change anything until there are auto commands that go over the charge station.

## Additional Notes

Currently, it checks if the auto command is a DefaultDriveCommand, which it won't be. This has to be changed to any commands that go over the charge station during auto. It does not have to trust vision more if it stops on the charge station, only if it passes over it to intake.


Kyle-Eldridge:
This is a very small PR. Can you merge it? It will make auto more accurate, so it should be on all of the auto branches.

src/main/java/frc/robot/Robot.java
```diff
@@ -124,6 +125,10 @@ public void autonomousInit() {
     if (m_autoCommand != null) {
       m_autoCommand.schedule();
     }
+
+    // Makes it trust vision more if it goes over the charge station
+    // TODO: Change this to whatever commands do that
+    m_robotContainer.setChargeStationAuto(m_autoCommand instanceof DefaultDriveCommand);
```
richtan, line 131: What if it is another command that isn't DefaultDriveComand but also isn't the charge station?

src/main/java/frc/robot/subsystems/Drivetrain.java
```diff
@@ -296,13 +299,20 @@ public void updateOdometry() {
...
-          estimatedPose.estimatedPose.toPose2d(),
-          estimatedPose.timestampSeconds,
-          VisionConstants.kBaseVisionPoseStdDevs.plus(
-            currentEstimatedPoseTranslation.getDistance(closestTagPoseTranslation) * VisionConstants.kVisionPoseStdDevFactor
-          )
-        );
+        if(m_chargeStationAuto){
+          m_poseEstimator.addVisionMeasurement(
+            estimatedPose.estimatedPose.toPose2d(),
+            estimatedPose.timestampSeconds
```
richtan, line 305: This just uses default std devs, it doesn't trust vision "more"

src/main/java/frc/robot/Robot.java
AnthonyFurman: I would just add this as part of the auto routine

src/main/java/frc/robot/subsystems/Drivetrain.java
AnthonyFurman: Should be switched i think

src/main/java/frc/robot/Robot.java
Kyle-Eldridge: This will be changed to something else, not DefaultDriveCommand. If it doesn't go over the charge station, don't add it here.

src/main/java/frc/robot/subsystems/Drivetrain.java
Kyle-Eldridge: It doesn't use any std devs, so doesn't that make it trust vision completely, like before you added them?

src/main/java/frc/robot/Robot.java
Kyle-Eldridge: I don't have the classes for the actual autos.

Kyle-Eldridge:
Change the command name in Robot.java when there are auto commands.

src/main/java/frc/robot/subsystems/Drivetrain.java
```diff
@@ -393,7 +373,12 @@ public void updateOdometry() {
 
     // Updates pose based on vision
     if (m_visionEnabled) {
-      //TODO: there should be a cleaner way to prosses vision
+
+      // If the cosine of the picth or roll is less than 1, it is not 0
+      // The cosine should be less than 0.995 (5.73 degrees) if it goes over the charge station
+      if(getPitch().getCos()<0.995 || getRoll().getCos()<0.995){
```
AnthonyFurman, line 379: Probably more clear to just check the angle itself rather than the cosine of the angle. Also, check spelling in the comments

src/main/java/frc/robot/subsystems/Drivetrain.java
```diff
@@ -422,19 +407,20 @@ public void updateOdometry() {
...
+          m_chargeStationVision ? VisionConstants.kChargeStationVisionPoseStdDevs : VisionConstants.kBaseVisionPoseStdDevs.plus(
             currentEstimatedPoseTranslation.getDistance(closestTagPoseTranslation) * VisionConstants.kVisionPoseStdDevFactor
           )
         );
       }
+      
+      // If it used vision after going over the charge station, it should trust vision normally again
+      if(estimatedPoses.size()>0){
+        m_chargeStationVision = false;
+      }
```
AnthonyFurman, line 419: Do you need to set it false only if there are vision targets present?
In fact, why is this a member variable anymore? You don't use it anywhere else right? You could just make it a local boolean

Kyle-Eldridge: I thought about that, but I want it to use vision 1 time after the angle is over that value. If it can't see an April tag while it is on the charge station, I want it to keep this variable as true until it can, so it corrects the pose after the charge station.

AnthonyFurman: Ok I see. Maybe make that more clear in comments.

AnthonyFurman (approved):

PR #43, opened 2023-02-27
Controller button updates

omegaladon:
## Overview
Remove instances of classes implementing Button, due to being deprecated (changed to Trigger).

### DIFF
--- a/src/main/java/lib/controllers/Ex3DProController.java
+++ b/src/main/java/lib/controllers/Ex3DProController.java
@@ -41,8 +41,8 @@ public class Ex3DProController extends Controller {
     }
   }
 
-  public JoystickButton get(Ex3DProButton button) {
-    return new JoystickButton(m_controller, button.id);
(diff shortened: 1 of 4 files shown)


src/main/java/lib/controllers/Ex3DProController.java
```diff
@@ -41,8 +41,8 @@ public enum Ex3DProAxis {
     }
   }
 
-  public JoystickButton get(Ex3DProButton button) {
-    return new JoystickButton(m_controller, button.id);
+  public Trigger get(Ex3DProButton button) {
+    return new Trigger(() -> m_controller.getRawButton(button.id));
```
PandaNinja10, line 45: I might be wrong about this but pretty sure the Ex3Dpro has a hat stick. You should add it in this PR as it is already pretty low priority so there is no rush.

PandaNinja10 (approved):

PR #44, opened 2023-02-27
Grid Alignment PR

Kyle-Eldridge:
## Overview

This PR will make the robot align to a grid spot selected by the operator when the driver holds a button.

## Design Doc

This doesn't have a design doc, and I don't have time to make one. The competition is this week. All this does is grid alignment.

## Tests Ran

I don't have the robot because other people are using it, so this is untested. It should work, and I can make any small changes in a separate PR.

## Additional Notes

Apparently, my Node class is being rewritten, so this might have to be changed to use that instead. For grid alignment, it needs to include where the robot should go to score, not just the Pose3d of the actual node. It should also use kGridDistance in VisionConstants.


Kyle-Eldridge:
I wanted this to make a PR now, while it is still a relatively small branch, even if it is untested.

src/main/java/frc/robot/commands/MoveToSelectedPose.java
```diff
@@ -0,0 +1,37 @@
...
+  private final Drivetrain m_drive;
+  
+  public MoveToSelectedPose(Drivetrain drive) {
+    m_drive = drive;    
+    addRequirements(drive);
+  }
+  
+  @Override
+  public void execute() {
+    m_drive.setAllOptimize(true);
```
AnthonyFurman, line 22: probably in init()

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -113,7 +112,8 @@ public RobotContainer() {
       m_intake = null;
       m_deployingBar = null;
 
-      m_operator = null;
+      m_operator = new Operator(m_drive, m_arm, m_vision, m_intake, m_deployingBar);
+      m_operator.configureControls();
```
PandaNinja10, line 116: update operator config fuctions and contructor to match main

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -137,7 +137,7 @@ public RobotContainer() {
     m_drive.setupModulesShuffleboard();
     m_vision.setupVisionShuffleboard();
     m_driver.setupShuffleboard();
-    m_vision.logging();
+    m_vision.logPose(m_drive.getPose());
```
PandaNinja10, line 140: This dose nothing but add a pose of (0,0,0) to our logs, and the robot pose logging should be in drivetrain

src/main/java/frc/robot/commands/MoveToSelectedPose.java
```diff
@@ -0,0 +1,40 @@
...
+  }
+  
+  @Override
+  public void execute() {
+    double x = Operator.selectedNode.scorePose.getX();
+    double y = Operator.selectedNode.scorePose.getY();
+    double rot = Operator.selectedNode.scorePose.getRotation().getRadians();
+
+    System.out.printf("Moving to (%.2f, %.2f) at %.2f degrees", x, y, Units.radiansToDegrees(rot));
```
PandaNinja10, line 31: don't have this in execute this will spam the rio log when ever it is called. if you really want to print something do so in initalize and end. Even then genrealy you shouldn't print stuff on normal commands.

src/main/java/frc/robot/commands/MoveToSelectedPose.java
```diff
@@ -0,0 +1,40 @@
+package frc.robot.commands;
```
PandaNinja10, line 1: This whole class could prorobly a subclass of 





























this command should probobly be a subclass of the GoToPoseCommand

src/main/java/frc/robot/commands/vision/TestVisionDistance.java
```diff
@@ -60,7 +71,6 @@ public void execute(){
       // reset the timer
       m_endTimer.reset();
       // If kPrintDelay seconds have passed, print the data
-      // TODO: Consider using shuffleboard instead of printing
```
PandaNinja10, line 63: is there a reason why you are not useing shuffleboard

src/main/java/frc/robot/constants/VisionConstants.java
```diff
@@ -39,8 +40,8 @@ public static void update() {
         new Pair<String, Transform3d>(
           "Camera_2",
           new Transform3d(
-            new Translation3d(-Units.inchesToMeters(4.75), Units.inchesToMeters(10.375), Units.inchesToMeters(10)),
-            new Rotation3d(0, 0, 0)
+            new Translation3d(Units.inchesToMeters(15), Units.inchesToMeters(-3.3125), Units.inchesToMeters(10)),
+            new Rotation3d(0, 0, Math.PI)
```
PandaNinja10, line 44: why is the rotation PI, the camera on the test bot is faceing the same direciton as the robot.

src/main/java/frc/robot/controls/Operator.java
```diff
@@ -1,37 +1,134 @@
...
+  private Vision m_vision;
+  private Intake m_intake;
+  private DeployingBar m_deployingBar;
+
+  // Values for selecting a node
+  // Grid, row, and column in the grid
+  private int[] selectValues = {1, 1, 1};
+
+  // The selected node where the robot will score
+  public static Node selectedNode = new Node();
```
PandaNinja10, line 41: generaly don't make viarables public or static.

src/main/java/frc/robot/controls/Operator.java
```diff
@@ -1,37 +1,134 @@
...
+
+  public Operator(Drivetrain drive, FourBarArm arm, Vision vision, Intake intake, DeployingBar deployingBar){
+    m_drive = drive;
+    m_arm = arm;
+    m_vision = vision;
+    m_intake = intake;
+    m_deployingBar = deployingBar;
+    m_operator = new GameController(OIConstants.kOperatorJoy);
+    selectValue(0, 1);
+  }
```
PandaNinja10, line 51: make operator match the config and contrutor standard on main

src/main/java/frc/robot/controls/Operator.java
```diff
@@ -1,37 +1,134 @@
...
   }
+
+  private final BooleanSupplier 
+    LEFT_STICK_LEFT = () -> m_operator.get(Axis.LEFT_X) < -0.75,
+    LEFT_STICK_RIGHT = () -> m_operator.get(Axis.LEFT_X) > 0.75,
+    LEFT_STICK_UP = () -> m_operator.get(Axis.LEFT_Y) < -0.75;
+  private final BooleanSupplier 
+    RIGHT_STICK_LEFT = () -> m_operator.get(Axis.RIGHT_X) < -0.75,
+    RIGHT_STICK_RIGHT = () -> m_operator.get(Axis.RIGHT_X) > 0.75,
+    RIGHT_STICK_UP = () -> m_operator.get(Axis.RIGHT_Y) < -0.75;
```
PandaNinja10, line 133: genrealy we diffine variables at the top

src/main/java/frc/robot/subsystems/Drivetrain.java
```diff
@@ -299,7 +299,7 @@ public void updateOdometry() {
         m_poseEstimator.addVisionMeasurement(
           estimatedPose.estimatedPose.toPose2d(),
           estimatedPose.timestampSeconds,
-          VisionConstants.kBaseVisionPoseStdDevs.plus(
+          VisionConstants.kBaseVisionPoseStdDevs.copy().plus(
```
PandaNinja10, line 302: sorry I brought this up the method makes a new object so this not needed

src/main/java/frc/robot/util/Vision.java
```diff
@@ -127,17 +135,16 @@ public Pose3d getTagPose(int id){
...
+  public void logPose(Pose2d referencePose){
+    Pose2d pose = getPose2d(referencePose);
+    if(pose!=null){
+      DoubleSupplier[] poseSupplier = {
+        () -> pose.getX(),
+        () -> pose.getY(),
+        () -> pose.getRotation().getRadians()
+      };
+      LogManager.addDoubleArray("Pose2d", poseSupplier);
+    }
```
PandaNinja10, line 147: this dosen't do anything but log one pose but as many entries and when you call it in robot container it will always be a (0,0,0) pose

PandaNinja10: also the pose should prorobly be loged in drivetrain

src/main/java/lib/controllers/GameController.java
```diff
@@ -85,6 +85,10 @@ public Trigger get(Button button) {
     return new Trigger(() -> m_controller.getRawButton(button.id));
   }
 
+  public Trigger get(BooleanSupplier booleanSupplier) {
+    return new Trigger(() -> booleanSupplier.getAsBoolean());
+  }
+
```
PandaNinja10, line 91: I don't get why this method you can just do "new Trigger(boleanSupplier)" the need to make method in game controller is riduclus. Similarly you don't need to convert a boolean supllier into a boolean supplier by makeing it a lambda.

src/main/java/frc/robot/util/Vision.java
```diff
@@ -67,7 +67,7 @@ public ArrayList<EstimatedRobotPose> getEstimatedPoses(Pose2d referencePose) {
     for (int i = 0; i < m_cameras.size(); i++) {
       Optional<EstimatedRobotPose> estimatedPose = m_cameras.get(i).getEstimatedPose(referencePose);
       // If the camera can see an april tag, add it to the array list
-      if (estimatedPose.isPresent()) {
+      if (estimatedPose.isPresent() && estimatedPose.get()!=null &&estimatedPose.get().estimatedPose!=null) {
```
PandaNinja10, line 70: do you really need to check that every bit is not a null, I am pretty sure if estimated Pose didn't exist than when you call it to get the pose it will spit out an error and "estimatedPose.isPresent()" and "estimatedPose.get()!=null" do the same thing.

src/main/java/frc/robot/commands/MoveToSelectedPose.java
Kyle-Eldridge: There is no GoToPose command on this branch. I think it's on an unmerged branch.

src/main/java/frc/robot/commands/vision/TestVisionDistance.java
Kyle-Eldridge: I'm not sure how to use shuffleboard. If it's similar to the logging in Vision, I don't know where it goes so I can read it.

src/main/java/frc/robot/constants/VisionConstants.java
Kyle-Eldridge: This is what worked on the classbot. 0 made the robot's rotation 0 when it faced toward red, and when I moved forward, it's pose moved in the negative x direction instead of positive x. This fixed it. π is facing toward red and 0 is facing toward blue.

src/main/java/frc/robot/controls/Operator.java
Kyle-Eldridge: I made it private, but I don't want to pass the operator into the driver, which I would have to do if it wasn't static.

src/main/java/frc/robot/util/Vision.java
Kyle-Eldridge: It used to log how many poses it was getting, but that doesn't help because the two cameras are always pointing the same direction, so it will almost always be 2. I did not write this, and whoever did thought it returned one pose for each April tag, which it doesn't. If you are using the comp bot, it will always say (0, 0) because the cameras are uncalibrated. This is vision pose, not robot pose, so it shouldn't be in Drivetrain.

src/main/java/lib/controllers/GameController.java
Kyle-Eldridge: It's easier to use m_operator.get() than new Trigger(), because that is what the rest of the code uses. The lambda is for getting as a boolean. All of the other methods in the class use it, too. This also used to exist, so I just re-added it.

src/main/java/frc/robot/util/Vision.java
Kyle-Eldridge: Sometimes, it tries to find April tags that don't exist, like 20, and I think those are present, but they don't return a valid pose.

Kyle-Eldridge: I also want to check if estimatedPose.get() is null because the next part of the line will crash the code if it's null, even if it should never be null.

src/main/java/frc/robot/RobotContainer.java
Kyle-Eldridge: Read comment below.

Kyle-Eldridge:
This should work now if you calibrate the cameras.

src/main/java/lib/controllers/GameController.java
chem7397: A Boolean supplier supplies a Boolean and therefore you don't need to get as a Boolean because it already does that and therefore a lambda is repetitive.

src/main/java/frc/robot/controls/Operator.java
```diff
@@ -80,17 +79,35 @@ public void configureControls(Wrist wrist, RollerIntake intake, Elevator elevato
...
+    tab.addStringArray("Selected node", ()->new String[]{
+      "Alliance: "+getSelectedNode().alliance,
+      "Type: "+getSelectedNode().type,
+      "Row: "+getSelectedNode().row,
+      "Column: "+getSelectedNode().column,
+      String.format("Score pose: (%.2f, %.2f) at %d degrees", 
+        getSelectedNode().scorePose.getX(), 
+        getSelectedNode().scorePose.getY(), 
+        getSelectedNode().scorePose.getRotation().getDegrees()
+      )
```
AnthonyFurman, line 109: why add a string array instead of several strings

src/main/java/frc/robot/util/NodeLegecy.java
```diff
@@ -0,0 +1,98 @@
...
+import edu.wpi.first.math.util.Units;
+import edu.wpi.first.wpilibj.DriverStation;
+import edu.wpi.first.wpilibj.DriverStation.Alliance;
+import frc.robot.constants.VisionConstants;
+
+/**
+ * Class to store data about scoring locations
+ * Stores location of node relative to driver's perspective
+*/
+public class NodeLegecy {
```
AnthonyFurman, line 16: what is this class for?

AnthonyFurman:
looks good, though for the std dev calculator, if you are going to be changing the array length when it's interrupted, and trying to append to the next empty spot in the array, then it would be easier to use an ArrayList. That makes the code easier to read and prevents unintended issues, I think there was an off-by-one error I found.
Also, please put spaces inbetween ) { and before/after operators like +=, ==, etc.

src/main/java/frc/robot/controls/GameControllerDriverConfig.java
```diff
@@ -32,6 +37,16 @@ public void configureControls() {
...
 
+    // Moves to the selected scoring position using Path Planner
+    kDriver.get(kDriver.LEFT_TRIGGER_BUTTON).whileTrue(new GoToNode(m_operator, getDrivetrain()));
+    // Moves to the selected scoring position using the PID
+    kDriver.get(Button.LB).whileTrue(new GoToNodePID(m_operator, getDrivetrain()));
+
+    // Moves to the shelf using Path Planner
+    kDriver.get(kDriver.RIGHT_TRIGGER_BUTTON).whileTrue(new GoToShelf(getDrivetrain()));
+    // Moves to the shelf using the PID
+    kDriver.get(Button.RB).whileTrue(new GoToShelfPID(getDrivetrain()));
```
AnthonyFurman, line 48: I think there are already buttons on the triggers for slowmode and angle align, for now I would rebind those to different buttons

src/main/java/frc/robot/controls/Operator.java
```diff
@@ -80,17 +79,35 @@ public void configureControls(Wrist wrist, RollerIntake intake, Elevator elevato
...
+    m_operator.get(m_operator.RIGHT_STICK_LEFT).onTrue(new InstantCommand(() -> selectValue(NodePositionIndex.COLUMN, 2)));
+    m_operator.get(m_operator.RIGHT_STICK_DOWN).onTrue(new InstantCommand(() -> selectValue(NodePositionIndex.COLUMN, 1)));
+    m_operator.get(m_operator.RIGHT_STICK_RIGHT).onTrue(new InstantCommand(() -> selectValue(NodePositionIndex.COLUMN, 0)));
     // Selects the row
-    // m_operator.get(DPad.LEFT).onTrue(new InstantCommand(() -> selectValue(NodePositionIndex.ROW, 2)));
-    // m_operator.get(DPad.DOWN).onTrue(new InstantCommand(() -> selectValue(NodePositionIndex.ROW, 1)));
-    // m_operator.get(DPad.RIGHT).onTrue(new InstantCommand(() -> selectValue(NodePositionIndex.ROW, 0)));
+    m_operator.get(Button.Y).onTrue(new InstantCommand(() -> selectValue(NodePositionIndex.ROW, 2)));
+    m_operator.get(Button.X).onTrue(new InstantCommand(() -> selectValue(NodePositionIndex.ROW, 1)));
+    m_operator.get(Button.A).onTrue(new InstantCommand(() -> selectValue(NodePositionIndex.ROW, 0)));
```
AnthonyFurman, line 92: Y, X, and A are used. Maybe talk with richie about what controls he wants or change the dpad controls to use a different button. I think the extra intake button is technically unnecessary, so that would free up a dpad button.

src/main/java/frc/robot/controls/GameControllerDriverConfig.java
Kyle-Eldridge: Where are those controls? I don't see them in the game controller class. These buttons are just for testing, but we probably will only use PathPlanner later and not the PID, so all alignment controls can be moved to the left side later.

src/main/java/frc/robot/controls/Operator.java
Kyle-Eldridge: Selecting the row doesn't really do anything except update Shuffleboard because the scoring position is still the same, so this can be on the same buttons as moving the arm/elevator.

src/main/java/frc/robot/commands/GoToNode.java
```diff
@@ -1,14 +1,80 @@
...
+    );
+    // Creates the command using the two points
+    m_command = new PathPlannerCommand(
+      new ArrayList<PathPoint>(List.of(point1, point2)),
+      m_drive,
+      false
+    );
+    // Starts the command
+    m_command.schedule();
+  }
```
PandaNinja10, line 60: this is really jank and prorobly isn't best practice. Because there are 2 commands this command could "run" but the path planner command fails to schedule resulting in this command never ending. This is a problem if this is used in a command group as the command group will sucessfuly schedule and and fail to run the PP command.

src/main/java/frc/robot/commands/GoToNodePID.java
```diff
@@ -0,0 +1,51 @@
...
+    m_drive.getXController().reset();
+    m_drive.getYController().reset();
+    m_drive.getRotationController().reset();
+  }
+  
+  @Override
+  public void execute() {
+    double x = m_operator.getSelectedNode().scorePose.getX();
+    double yOffset = (DriverStation.getAlliance() == Alliance.Blue?1:-1)*0; //TODO: Add distance sensor measurement
+    double y = m_operator.getSelectedNode().scorePose.getY() + yOffset;
```
PandaNinja10, line 34: shouldn't the off set be in the node score pose

src/main/java/frc/robot/commands/GoToNode.java
Kyle-Eldridge: How should I do it instead?

Kyle-Eldridge: @PandaNinja10

PandaNinja10: The best extra percation I could suggest at the momment is adding the .isScheduled() to the is finished so if the command fails to schedule at least the command will end. Sorry, it is a comment on a problem I have no easy solution to a the momnet.

PandaNinja10: I could suggest some weird janky ways to use sequential command groups but it is complealy untested and last time I tried to do someting like an error was spat out.

PandaNinja10: OK so I did some testing and my new method of jankyness dosen't spit out an error as far as I am aware so here is the code:

public class SequentialExploit extends SequentialCommandGroup {
  
  public SequentialExploit(){
    SequentialCommandGroup commandBuffer = new SequentialCommandGroup();
    addCommands(new InstantCommand(() -> {command.addCommands(commandToBecratedOnInitalization);}),commandBuffer);
  }
}

PandaNinja10: The only I ssue I am aware of is that you need ot add the requirement in the root command by hand as the requirement is not there initaly

src/main/java/frc/robot/commands/GoToShelf.java
```diff
@@ -0,0 +1,75 @@
...
+    );
+    // Creates the command using the two points
+    m_command = new PathPlannerCommand(
+      new ArrayList<PathPoint>(List.of(point1, point2)),
+      m_drive,
+      false
+    );
+    // Starts the command
+    m_command.schedule();
+  }
```
PandaNinja10, line 55: this is really jank and prorobly isn't best practice. Because there are 2 commands this command could "run" but the path planner command fails to schedule resulting in this command never ending. This is a problem if this is used in a command group as the command group will sucessfuly schedule and and fail to run the PP command.

src/main/java/frc/robot/commands/auto/PathPlannerCommand.java
```diff
@@ -25,12 +25,15 @@ public class PathPlannerCommand extends SequentialCommandGroup {
...
+
+  public PathPlannerCommand(ArrayList<PathPoint> waypoints, Drivetrain drive, boolean useAllianceColor) {
     this(new ArrayList<PathPlannerTrajectory>(Arrays.asList(PathPlanner.generatePath(
       new PathConstraints(AutoConstants.kMaxAutoSpeed, AutoConstants.kMaxAutoAccel),
       waypoints.get(0),
-      waypoints.get(1),
-      (PathPoint[]) Arrays.copyOfRange(waypoints.toArray(), 2, waypoints.size())
-    ))), 0, drive, true);
+      waypoints.get(1)
+    ))), 0, drive, false, useAllianceColor);
```
PandaNinja10, line 36: is there a reson why reset pose is not true by defult like it was before?

src/main/java/frc/robot/commands/vision/CalculateStdDevs.java
```diff
@@ -0,0 +1,139 @@
...
+   * @param time How long to run the command, also how many poses to use
+   * @param shuffleboardTab The shuffleboard tab to output to
+   * @param drive The drivetrain
+   * @param vision The vision
+   */
+  public CalculateStdDevs(int time, ShuffleboardTab shuffleboardTab, Drivetrain drive, Vision vision) {
+    m_tab = shuffleboardTab;
+    m_drive = drive;
+    m_vision = vision;
+    m_arrayLength = time;
```
PandaNinja10, line 38: I think "posesToUse" is likely a better name for "time" as the time it runs can be determined by noise and is unclear for what unit it would be in.

src/main/java/frc/robot/commands/vision/CalculateStdDevs.java
```diff
@@ -0,0 +1,139 @@
...
+    if (pose != null) {
+      m_endTimer.stop();
+      m_endTimer.reset();
+      m_poses.add(pose);
+      System.out.printf("%.1f%% done", ((double)m_poses.size())/m_arrayLength * 100);
+    } else {
+      m_endTimer.start();
+      // If 5 seconds have passed since it saw an April tag, stop the command
+      // Prevents it from running forever
+      if(m_endTimer.hasElapsed(5)){
```
PandaNinja10, line 65: maybe a constant for how long it takes.

src/main/java/frc/robot/commands/vision/CalculateStdDevs.java
```diff
@@ -0,0 +1,139 @@
...
+    double totalY = 0;
+    double totalRot = 0;
+    for (int i = 0; i < m_poses.size(); i++) {
+      totalX += Math.pow(m_poses.get(i).getX() - meanX, 2);
+      totalY += Math.pow(m_poses.get(i).getY() - meanY, 2);
+      totalRot += Math.pow(m_poses.get(i).getRotation().getRadians() - meanRot, 2);
+    }
+    double stdDevX = Math.sqrt(totalX / m_poses.size());
+    double stdDevY = Math.sqrt(totalY / m_poses.size());
+    double stdDevRot = Math.sqrt(totalRot / m_poses.size());
```
PandaNinja10, line 104: it should be : Math.sqrt(total / (m_poses.size()-1))
because it is a sample and not a population

src/main/java/frc/robot/commands/vision/CalculateStdDevs.java
```diff
@@ -0,0 +1,139 @@
...
+    // Calculate distance to closest April tag
+    double closest = Double.POSITIVE_INFINITY;
+    ArrayList<EstimatedRobotPose> estimatedPoses = m_vision.getEstimatedPoses(m_drive.getPose());
+    for (int i = 0; i < estimatedPoses.size(); i++) {
+      for (int j = 0; j < estimatedPoses.get(i).targetsUsed.size(); j++) {
+        double distance = estimatedPoses.get(i).targetsUsed.get(j).getBestCameraToTarget()
+        .getTranslation().toTranslation2d().getNorm();
+        closest = Math.min(closest, distance);
+      }
+    }
```
PandaNinja10, line 115: I think it might be a good idea to make a statisitics util class so if we ever want to use this for other things

src/main/java/frc/robot/commands/vision/CalculateStdDevs.java
```diff
@@ -0,0 +1,139 @@
...
+    
+    // Store data in an array to make logging and shuffleboard easier
+    double[] stdDevs = {stdDevX, stdDevY, stdDevRot, closest};
```
PandaNinja10, line 118: First off it is not easier to log and dissplay on shuffleboard.  second the closest distance is not even a standard deviation so it shouldn't be in avariable called stdDev.

src/main/java/frc/robot/commands/vision/CalculateStdDevs.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+    
+    // Calculate std devs
+    double meanX = 0;
+    double meanY = 0;
+    double meanRot = 0;
+    for (int i = 0; i < m_poses.size(); i++) {
+      meanX += m_poses.get(i).getX();
+      meanY += m_poses.get(i).getY();
+      meanRot += m_poses.get(i).getRotation().getRadians();
```
vfurman-gh, line 89: It is wrong to compute mean rotation by averaging angle. Average of 0 and 360 degrees is 0 (or 360), but using this logic it will be 180. A simple way to do it correctly is to operate on Rotation objects - sum them up and divide, and then take the angle

src/main/java/frc/robot/constants/VisionConstants.java
```diff
@@ -34,29 +45,29 @@ public static void update(RobotId robotID) {
...
-
-      //   new Pair<String, Transform3d>(
-      //     "Camera_2",
-      //     new Transform3d(
-      //       new Translation3d(Units.inchesToMeters(1.25), Units.inchesToMeters(-7.125), Units.inchesToMeters(21)),
-      //       new Rotation3d(0, 0, 0)
-      //     )
-      //   )
-      // ));
+      // DO NOT COMMENT THIS OUT
```
PandaNinja10, line 54: I think the comment is unnecessary

src/main/java/frc/robot/util/Node.java
```diff
@@ -39,9 +42,9 @@ public Node()  {
    * @param row
    *  row it's in (1 = bottom, 2 = middle, 3 = top)
    * @param column
-   *  column from field boundary to other boundary (1 to 9)
+   *  column from field boundary to loading zone (1 to 9)
    */
-  public Node(Alliance alliance, int row, int column) {
+  public Node(RollerIntake intake, Alliance alliance, int row, int column) {
```
PandaNinja10, line 47: Personaly I think the node class should be held in the node class. If the off set is going to be handled int he node class I think the game pice offset should be a method parameter for geting the pose.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -94,13 +96,14 @@ public RobotContainer(RobotId robotId) {
 
         // Create Drivetrain
         m_drive = new Drivetrain(m_drivetrainTab, m_swerveModulesTab, m_vision);
-        m_driver = new GameControllerDriverConfig(m_drive, m_controllerTab, false);
+        m_operator = new Operator();
+
+        m_driver = new GameControllerDriverConfig(m_drive, m_operator, m_controllerTab, false);
```
AnthonyFurman, line 101: fairly minor but consider just giving the driver a supplier for the node. It's kinda weird the driver controller "has" the operator controller

src/main/java/frc/robot/commands/GoToNodePID.java
```diff
@@ -0,0 +1,48 @@
...
+    m_drive.getXController().reset();
+    m_drive.getYController().reset();
+    m_drive.getRotationController().reset();
+  }
+  
+  @Override
+  public void execute() {
+    double x = m_operator.getSelectedNode().scorePose.getX();
+    double y = m_operator.getSelectedNode().scorePose.getY();
+    double rot = m_operator.getSelectedNode().scorePose.getRotation().getRadians();
```
AnthonyFurman, line 32: for safety, and making it more readible, you can make a m_scorePose set to to the operator's control in init().

src/main/java/frc/robot/commands/auto/PathPlannerCommand.java
AnthonyFurman: Well if you are constructing points from your current pose no reason to reset pose to your current pose, that does nothing. And if your pose happens to change might be resetting to the wrong pose. The only time you should really reset your pose is if you are sure you are at the location, which only happens in auto (the first path)

src/main/java/frc/robot/commands/vision/CalculateStdDevs.java
AnthonyFurman: I agree

src/main/java/frc/robot/commands/vision/CalculateStdDevs.java
```diff
@@ -0,0 +1,151 @@
...
+    
+    // Store data in an array to make logging and shuffleboard easier
+    double[] stdDevs = {stdDevX, stdDevY, stdDevRot, closest};
+    
+    // Print, log, and add the values to Shuffleboard
+    System.out.printf("Standard deviation values:\nX: %.5f\nY: %.5f\nRotation: %.5f\nDistance: %.5f\n",
+    stdDevX, stdDevY, stdDevRot, closest);
+    if (Constants.kLogging) {
+      LogManager.addDoubleArray("Vision/StdDevs", stdDevs);
+    }
```
AnthonyFurman, line 140: for some reason the shuffleboard display was crashing so I removed it.

src/main/java/frc/robot/commands/vision/CalculateStdDevs.java
PandaNinja10: I would like to point out that this will still result is avrageing issues. Example, 5 messuements of 90 degrees added together is 90 degreess then diveded by 5 is 18 degrees wich is clearly not the mean. Additionaly this is only an issue if the pose outputs a values around the boarder of the continous. So if the output is -pi to pi a mean of 0 won't cause issues and if it is 0 to 2pi then a mean of pi won't cause issues.

PandaNinja10: I would suggest adding documentation to state wich aprial tags could be used for the test.

src/main/java/frc/robot/commands/vision/CalculateStdDevs.java
```diff
@@ -0,0 +1,151 @@
...
+  public void initialize() {
+    m_poses = new ArrayList<Pose2d>();
+  }
+
+  /**
+   * Adds a pose to the array
+   */
+  @Override
+  public void execute() {
+    Pose2d pose = m_vision.getPose2d(m_drive.getPose());
```
PandaNinja10, line 55: just a thought but wouldn't the raw transformation from the camera be a better data point then the posibly inacurate odometry position, epecialy if you don't disable vision from effecting the pose estimator.

src/main/java/frc/robot/commands/GoToNode.java
Kyle-Eldridge: Wouldn't this add the same command to the other group every time this is run? If it is stored as a private variable, it doesn't get reset when the command is run a second time.

Kyle-Eldridge: Maybe I can add a resetCommand function after the command group.

Kyle-Eldridge: Currently, I can't find a way to reset it without recreating it. If I did that, would it run the new command or the old one? One complicated fix is to copy and paste the SequentialCommandGroup code into a new command and make some variables, like m_commands, public so we can remove index 1.

src/main/java/frc/robot/commands/GoToShelf.java
Kyle-Eldridge: I don't think we are going to use this in a command group this year. Do you know how other teams programmed this to work?

src/main/java/frc/robot/commands/vision/CalculateStdDevs.java
Kyle-Eldridge: I thought it was the square root of the average oft he difference squared. To take the average, it adds up all of the values and divides by the number of values. Why would it be -1 for a sample?

src/main/java/frc/robot/commands/vision/CalculateStdDevs.java
Kyle-Eldridge: It might be easier on some computers where printing doesn't work correctly on the rio log on VSCode, like mine. On DriverStation, it sometimes covers important prints with warnings and other prints, and I don't want to accidentally lose the data. Also, the distance is related to the std dev test, in case it changes at different distances.

src/main/java/frc/robot/commands/vision/CalculateStdDevs.java
Kyle-Eldridge: I mentioned on Slack to only use red April tags because of the averaging, because the angle won't be near -π or π.

src/main/java/frc/robot/constants/VisionConstants.java
Kyle-Eldridge: It is necessary. This was commented out during SFR instead of disabling using the variable. I don't want it to be commented out again.

src/main/java/frc/robot/RobotContainer.java
Kyle-Eldridge: I originally had selectedNode as public so I wouldn't have to do that, but I had to change it to private. I think a supplier might make it even more confusing. (using m_operator.getSelectedNodeSupplier() instead of m_operator).

src/main/java/frc/robot/commands/GoToNodePID.java
Kyle-Eldridge: If it was going in the wrong direction, I wanted the operator to be able to change the selected node and make the robot go there instead. I didn't add it to the other GoToNode because the PathPlanner command can't be recreated every frame.

src/main/java/frc/robot/commands/vision/CalculateStdDevs.java
Kyle-Eldridge: This uses Vision.getPose2d(), so it doesn't use the drivetrain. I also used the drivetrain as a reference and didn't disable vision to simulate how the robot will use vision on the field.

src/main/java/frc/robot/commands/vision/CalculateStdDevs.java
Kyle-Eldridge: Also, when this comment means is that the array makes it easier to log and use Shuffleboard instead of creating it in the same line.

src/main/java/frc/robot/commands/vision/CalculateStdDevs.java
Kyle-Eldridge: After Houston, I'll add more documentation, but there are other things to do until then.

src/main/java/frc/robot/commands/GoToNode.java
PandaNinja10: Nevermind the way listed above only works if the command object is called once, wich I forgot is not the case in your instance. You can just add the saftey check that the command actualy scheduled unless we find a way to work around path planner being hard to work with.

Kyle-Eldridge: Do you only want it to end if the command is scheduled or only if it isn't? Once it finishes, I don't think it's scheduled anymore.

src/main/java/frc/robot/constants/VisionConstants.java
Kyle-Eldridge: image
This is why it is necessary.

AnthonyFurman: There are reasons to remove the cameras.
For example if mechanical removes a camera.
Another example is the sim, which doesn't support the cameras (though there should be some code to support that).
In general, writing a comment like "don't comment this out" is strange because there are cases where you want to disable things. Commenting is a way of writing code, it would be like writing "don't write code". Now there are some cases where maybe some code is very crucial and should not be modified at all (like Main.java, which has a comment), but that's not the case for VisionConstants.java

src/main/java/frc/robot/commands/GoToPosePID.java
```diff
@@ -0,0 +1,46 @@
...
+    addRequirements(drive);
+}
+  
+@Override
+public void initialize() {
+}
+
+@Override
+public void execute() {
+  m_drive.runChassisPID(m_pose.get().getX(), m_pose.get().getY(), m_pose.get().getRotation().getRadians());
```
PandaNinja10, line 33: there will be a method in the drivetrain cleanup PR that acepts a pose that should make this line of code much more compact and easier to fallow. Just something to note when the PR get merged into main.

src/main/java/frc/robot/constants/VisionConstants.java
```diff
@@ -3,25 +3,41 @@
...
+  public static final double kChargeStationAngle = 2.5;
+
+  //TODO: Change these to whatever the actual distances are
+  // How far from the grid the robot should be to score for alignment in meters
+  public static double kGridDistanceAlignment = 0 + DriveConstants.kRobotWidthWithBumpers / 2; // meters
+  // How far from the shelf the robot should be to intake for alignment in meters
+  public static double kShelfDistanceAlignment = 0.5 + DriveConstants.kRobotWidthWithBumpers / 2;
+
+  public static Pose2d kBlueShelfAlignPose = new Pose2d(FieldConstants.kBlueShelfX - kShelfDistanceAlignment, FieldConstants.kShelfY, new Rotation2d(0));
+  public static Pose2d kRedShelfAlignPose = new Pose2d(FieldConstants.kRedShelfX + kShelfDistanceAlignment, FieldConstants.kShelfY, new Rotation2d(Math.PI));
```
PandaNinja10, line 38: I think these are better fit for the feild constansts  but it is up for debate

src/main/java/frc/robot/constants/IntakeConstants.java
```diff
@@ -13,7 +13,7 @@ public final class IntakeConstants {
   // intake speeds
   public static final double kIntakeCubePower = 0.6;
   public static final double kOuttakeCubePower = -0.6;
-  public static final double kIntakeConePower = -0.6;
+  public static final double kIntakeConePower = -0.75;
```
PandaNinja10, line 16: separate PR

src/main/java/frc/robot/constants/ElevatorConstants.java
```diff
@@ -83,7 +83,7 @@ public class ElevatorConstants {
     public static final double kBottomConeHeight = Conversions.ElevatorExtensionToHeight(Units.inchesToMeters(6));
     public static final double kBottomCubeHeight = kBottomConeHeight;
     public static final double kShelfHeight = 1.30;
-    public static final double kStowHeight = Conversions.ElevatorExtensionToHeight(Units.inchesToMeters(0));
+    public static final double kStowHeight = Conversions.ElevatorExtensionToHeight(Units.inchesToMeters(2));
```
PandaNinja10, line 86: separate PR

PandaNinja10: Anthony: "remove that"

src/main/java/frc/robot/constants/WristConstants.java
```diff
@@ -13,7 +13,7 @@ public class WristConstants {
   public static final int kPeakCurrentLimit = 45;
   public static final double kPeakCurrentDuration = 0.5;
 
-  public static final double kP = 2;
+  public static final double kP = 1.9;
```
PandaNinja10, line 16: separate PR

src/main/java/frc/robot/controls/Operator.java
```diff
@@ -30,15 +32,17 @@ public class Operator {
...
+
+  private RollerIntake m_intake;
   
   /**
    * Configures the operator controls for the wrist, roller intake, elevator, and vision
    */
   public void configureControls(Wrist wrist, RollerIntake intake, Elevator elevator, Vision vision) {
-    m_vision = vision; //should be in constructor
+    m_intake = intake;
+    selectValue(NodePositionIndex.ROW, 3);
```
PandaNinja10, line 41: seams a bit random

PandaNinja10: add a comment on why you have it there

src/main/java/frc/robot/util/Vision.java
```diff
@@ -98,23 +103,32 @@ public Pose2d getPose2d(Pose2d... referencePoses){
...
-        ))
+        estimatedPoses.get(0).estimatedPose.toPose2d().getTranslation()
+          .plus(estimatedPoses.get(1).estimatedPose.toPose2d().getTranslation())
+          .div(2),
+          
+          new Rotation2d(Functions.modulusMidpoint(
+            estimatedPoses.get(0).estimatedPose.toPose2d().getRotation().getRadians(),
+            estimatedPoses.get(1).estimatedPose.toPose2d().getRotation().getRadians(),
+            -Math.PI, Math.PI
+          )
```
PandaNinja10, line 119: this can be intent one less

src/main/java/frc/robot/commands/vision/CalculateStdDevs.java
```diff
@@ -0,0 +1,127 @@
...
+    ArrayList<EstimatedRobotPose> estimatedPoses = m_vision.getEstimatedPoses(m_drive.getPose());
+    for (int i = 0; i < estimatedPoses.size(); i++) {
+      for (int j = 0; j < estimatedPoses.get(i).targetsUsed.size(); j++) {
+        double distance = estimatedPoses.get(i).targetsUsed.get(j).getBestCameraToTarget()
+          .getTranslation().toTranslation2d().getNorm();
+        closest = Math.min(closest, distance);
+      }
+    }
+    
+    // Print, log, and add the values to Shuffleboard
```
PandaNinja10, line 108: values are not added to suffleboard

src/main/java/frc/robot/util/StatisticsUtil.java
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  public static double stdDev(ArrayList<Double> data){
+    return stdDev(arrayListToArray(data));
+  }
+
+  public static double stdDev(double[] data){
+    if(data.length == 0){
+      return 0;
+    }
```
PandaNinja10, line 34: spaces and you could put this on one line

PandaNinja10: also may want to inclued 1 as 1-1=0 so you will divide by 0.

src/main/java/frc/robot/commands/vision/CalculateStdDevs.java
PandaNinja10: updated comment to be more acurate

PandaNinja10:
ok once drivetrain clean up and Supplier command are merged in to main and we update the thing reliying on it it this PR is ready to go.

AnthonyFurman:
Looks good, after a long wait... approved!

In just under 100 comments and 170 commits...

AnthonyFurman (approved):

PR #47, opened 2023-02-27
Camera Positions PR

Kyle-Eldridge:
## Overview

This PR will add camera constants to the robot.

## Design Doc

Same as vision design doc.
https://docs.google.com/document/d/1UfKGZ_5jCUP1P7YjEF6BluMhBr28Fxi27rlz2LOM77A/edit

## Tests Ran

None. I don't have the robot. These might need to be changed if it isn't accurate.

## Additional Notes

Based on what I saw on the classbot 2, these should be accurate, but I don't know if there's a difference for swerve.

### DIFF
--- a/src/main/java/frc/robot/constants/VisionConstants.java
+++ b/src/main/java/frc/robot/constants/VisionConstants.java
@@ -44,6 +44,24 @@ public class VisionConstants {
           )
         )
       ));
+    }else{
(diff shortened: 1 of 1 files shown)


richtan (commented):
Maybe you should get the values from CAD instead

AnthonyFurman (approved):

PR #54, opened 2023-03-02
Vision fixes

Kyle-Eldridge:
## Overview
Fixes vision.


Kyle-Eldridge:
MERGE THIS. It could fix odometry errors in auto. This only changes the constants, adds and extra check for null poses, and adds vision logging. If the problem in auto today was not odometry / vision, this won't fix it, but it also won't make it worse.

PR #60, opened 2023-03-09
Basic swerve sim

AnthonyFurman:
## Overview

Create swerve simulation. Useful for testing the autos.

## Design Doc

See simulation section


src/main/java/frc/robot/Robot.java
```diff
@@ -72,6 +72,10 @@ public void robotInit() {
     // report the RobotId to the SmartDashboard
     SmartDashboard.putString("RobotID", kRobotId.name());
 
+    if (isSimulation()) {
+      kRobotId = RobotId.SwerveCompetition;
+    }
+
```
GLRoylance, line 78: This should not be needed. Your simulation can set the RobotId preference. 

I made RobotId more abstract in a new PR.

vfurman-gh:
I just added a commit that makes Sim Auto working. I think we can push this in main

src/main/java/frc/robot/Robot.java
omegaladon: Just curious, how would the simulation set the RobotId preference?

src/main/java/frc/robot/subsystems/Drivetrain.java
```diff
@@ -543,6 +552,7 @@ private void setupDrivetrainShuffleboard() {
     if (!Constants.kUseTelemetry) return;
 
     m_drivetrainTab.add("Field", m_fieldDisplay);
+    SmartDashboard.putData("Field Display", m_fieldDisplay);
```
omegaladon, line 555: We shouldn't need this if the field is already being put onto the drivetrain tab. You can view the field by using the field on the drivetrain tab, both in the simgui and in shuffleboard.

src/main/java/frc/robot/Robot.java
omegaladon: Never mind me, didn't read the comment in the code

src/main/java/frc/robot/commands/auto/PathPlannerCommand.java
```diff
@@ -50,20 +48,20 @@ public PathPlannerCommand(List<PathPlannerTrajectory> pathGroup, int pathIndex,
...
                 path, 
                 drive::getPose, // Pose supplier
-                DriveConstants.kKinematics, // SwerveDriveKinematics
                 drive.getPathplannerXController(), // X controller can't normal PID as pathplanner has Feed Forward 
                 drive.getPathplannerYController(), // Y controller can't normal PID as pathplanner has Feed Forward 
                 drive.getPathplannerRotationController(), // Rotation controller can't normal PID as pathplanner has Feed Forward 
-                drive::setModuleStates, // Module states consumer
-                true,
+                (chassisSpeeds) -> { drive.setChassisSpeeds(chassisSpeeds, false); }, // chassis Speeds consumer
+                false,  // use Alliance color
```
PandaNinja10, line 64: remove the path preseting and revert the use aliance color to true or else it will brake auto paths in main if we switch teams between enables.

src/main/java/frc/robot/subsystems/Drivetrain.java
```diff
@@ -336,6 +341,10 @@ public SwerveModuleState[] getModuleStates() {
    * @param isOpenLoop if open loop control should be used for the drive velocity
    */
   public void setChassisSpeeds(ChassisSpeeds chassisSpeeds, boolean isOpenLoop) {
+    if (!RobotBase.isReal()) {
```
PandaNinja10, line 344: just a small note there is a "Robot.isSimulation()" I think might be slightly more reable than "!RobotBase.isReal()"

src/main/java/frc/robot/subsystems/Intake.java
```diff
@@ -9,9 +9,7 @@
```
PandaNinja10, line 9: why are the distance sensors removed entierely rather than removed when robot is a simulation

src/main/deploy/pathplanner/MyTestPath.path
```diff
@@ -0,0 +1,74 @@
+{
```
PandaNinja10, line 1: better path name or remove

PandaNinja10 (approved):

PR #74, opened 2023-03-12
Roller intake preserved

omegaladon:
Adds roller intake code. 

Nothing is tested on an actual robot.


AnthonyFurman:
tested on robot, except current values

src/main/java/frc/robot/commands/scoring/Stow.java
```diff
@@ -1,22 +1,22 @@
...
 
 public class Stow extends SequentialCommandGroup {
-  public Stow(Intake intake, Elevator elevator, FourBarArm arm) {
+  public Stow(RollerIntake intake, Elevator elevator, FourBarArm arm) {
     addRequirements(intake, elevator, arm);
     addCommands(
-      new InstantCommand(() -> intake.setIdleMode(IdleMode.kBrake)),
+      new InstantCommand(() -> intake.setNeutralMode(NeutralMode.Brake)),
       new ExtendArm(arm, ArmConstants.kStowPos),
       new MoveElevator(elevator, ElevatorConstants.kStowHeight)
```
AnthonyFurman, line 21: I don't think this is necessary anymore, we don't need to set the intake to brake mode anymore so instead it can move as part of PositionRollerIntake

src/main/java/frc/robot/commands/scoring/intake/IntakeGamePiece.java
```diff
@@ -2,34 +2,48 @@
...
 
 public class IntakeGamePiece extends CommandBase {
-  Intake m_intake; 
+
+  private final RollerIntake m_intake; 
+  private final IntakePiece m_type;
 
   /**
    * Spins the intake until the game piece is inside the intake.
    * @param intake the intake subsystem
```
AnthonyFurman, line 16: update javadoc with parameters, doesn't have IntakePiece type

src/main/java/frc/robot/commands/scoring/intake/IntakeGamePiece.java
```diff
@@ -2,34 +2,48 @@
...
   
-  // Returns true when the command should end.
   @Override
   public boolean isFinished() {
-    return false;//m_intake.containsGamePiece(); 
+    if (m_type == IntakePiece.CUBE) {
+      return Math.abs(m_intake.getStatorCurrent()) >= IntakeConstants.kCubeIntakeCurrentStopPoint;
+    } else if (m_type == IntakePiece.CONE) {
+      return Math.abs(m_intake.getStatorCurrent()) >= IntakeConstants.kConeIntakeCurrentStopPoint;
+    }
```
AnthonyFurman, line 45: checking the currents like this probably won't work. The current spikes for an instant when first spinning up the motors, so that spike would trigger this. It probably needs to be above the current for a certain duration.

src/main/java/frc/robot/commands/scoring/intake/IntakeGamePiece.java
```diff
@@ -2,34 +2,48 @@
...
+      m_intake.setIntakeMode(IntakeMode.INTAKE_CUBE);
+    } else if (m_type == IntakePiece.CONE) {
+      m_intake.setIntakeMode(IntakeMode.INTAKE_CONE);
+    }
   }
 
   @Override
   public void end(boolean interrupted) {
-    m_intake.stopIntake(); 
+    if (!interrupted) m_intake.setHeldGamePiece(m_type);
```
AnthonyFurman, line 35: this should probably happen even if it is interrupted, because the driver can interrupt by releasing the controller earlier.
Even if it is interrupted and does set an that it has a game piece, it wouldn't matter because the robot would have to intake another game piece, setting it to the right mode, before outtaking.

src/main/java/frc/robot/commands/scoring/intake/OuttakeGamePiece.java
```diff
@@ -0,0 +1,60 @@
...
+
+  @Override
+  public void initialize() {
+    if (m_heldPiece == IntakePiece.CUBE) {
+      m_intake.setMode(IntakeMode.OUTTAKE_CUBE);
+    } else if (m_heldPiece == IntakePiece.CONE) {
+      m_intake.setMode(IntakeMode.OUTTAKE_CONE);
+    } else {
+      cancel();
+      return;
```
AnthonyFurman, line 44: just to be safe, maybe it should still outtake but give some sort of error? in case it thinks it doesn't have a game piece but it does.

src/main/java/frc/robot/subsystems/RollerIntake.java
```diff
@@ -0,0 +1,125 @@
...
+  public boolean containsCone() {
+    return m_heldPiece == IntakePiece.CONE;
+  }
+
+  public boolean containsCube() {
+    return m_heldPiece == IntakePiece.CUBE;
+  }
+
+  public double getStatorCurrent() {
+    return Math.abs(m_intakeMotor.getStatorCurrent());
```
AnthonyFurman, line 111: is there a reason you take the absolute value? Also, I think you should use supply current, as that is the current drawn by the motor, not the current provided in the motor controller

src/main/java/frc/robot/subsystems/RollerIntake.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  public void setNeutralMode(NeutralMode mode) {
+    m_intakeMotor.setNeutralMode(mode);
+  }
```
AnthonyFurman, line 116: is there a reason to change the neutral mode of the motors so often?

src/main/java/frc/robot/commands/scoring/intake/IntakeGamePiece.java
omegaladon: I added a check for if some duration had passed since starting the command. Would that work?

src/main/java/frc/robot/commands/scoring/intake/OuttakeGamePiece.java
omegaladon: For which game piece should it outtake for? Just outtaking isn't going to work because cubes have a negative outtake value.

AnthonyFurman (approved):

PR #78, opened 2023-03-13
Update staged game piece

omegaladon:
Makes StagedGamePiece give a translation2d instead of a pose2d. Also adds gamepiecetype


omegaladon:
ideally this can be merged before roller intake, so we can use the enum GamePieceType in roller intake code.

AnthonyFurman (approved):

PR #86, opened 2023-03-16
Drivetrain cleanup

PandaNinja10:
cleaning drivetrain and module code to make it more readable


src/main/java/frc/robot/subsystems/Drivetrain.java
```diff
@@ -54,60 +55,61 @@
...
     m_yPosEntry,
     m_headingEntry;
-  
+
   private Double[] m_driveVelFeedForwardSaver = new Double[4];
   private Double[] m_driveStaticFeedForwardSaver = new Double[4];
   private Double[] m_steerVelFeedForwardSaver = new Double[4];
   private Double[] m_steerStaticFeedForwardSaver = new Double[4];
-  
+
```
AnthonyFurman, line 102: As a great man once said, "Remove random spaces, I know it is knit picky but it is annoying for a few reasons. It make the file change view larger than it needs to be. It makes random files appear in the pull request. Random spaces can add up and It makes random gaps of code which increases the distance to scroll though code. Everyone dose it but at least clean it up before you make a PR. So I know it's knit picky but it makes is easier for the people reviewing and working on the code."

src/main/java/frc/robot/subsystems/Drivetrain.java
```diff
@@ -407,41 +279,95 @@ public void updateOdometry() {
...
+  }
+
   /**
    * Drives the robot using the provided x speed, y speed, and positional heading.
    * 
    * @param xSpeed speed of the robot in the x direction (forward)
    * @param ySpeed speed of the robot in the y direction (sideways)
    * @param heading target heading of the robot in radians
-   * @param fieldRelative whether the provided x and y speeds are relative to the field
+   * @param  whether the provided x and y speeds are relative to the field
```
AnthonyFurman, line 357: why did you remove the title of the param

src/main/java/frc/robot/subsystems/Drivetrain.java
```diff
@@ -454,89 +380,204 @@ public void runChassisPID(double x, double y, double rot) {
...
+    SwerveDriveKinematics.desaturateWheelSpeeds(swerveModuleStates, DriveConstants.kMaxSpeed);
+    for (int i = 0; i < 4; i++) {
+      m_modules[i].setDesiredState(swerveModuleStates[i], isOpenLoop);
+    }
+  }
 
-    // outputs in deg/s, so convert to rad/s
-    return Units.degreesToRadians(rawGyros[id]);
+  /**
+   * Sets the desired states for all swerve modules. Runs closed loop control. USe
```
AnthonyFurman, line 408: "USe" check capitalization

src/main/java/frc/robot/subsystems/Drivetrain.java
```diff
@@ -454,89 +380,204 @@ public void runChassisPID(double x, double y, double rot) {
...
+   * @return chassis speed of swerve drive
+   */
+  public ChassisSpeeds getChassisSpeeds() {
+    return DriveConstants.kKinematics.toChassisSpeeds(getModuleStates());
+  }
+
+  public ChassisSpeeds getFieldRelativeChassisSpeeds() {
+    return ChassisSpeeds.fromFieldRelativeSpeeds(
+      getChassisSpeeds(),
+      getPose().getRotation());
```
AnthonyFurman, line 523: getYaw() could be more clear

src/main/java/frc/robot/subsystems/Module.java
```diff
@@ -66,106 +64,29 @@ public Module(ModuleConstants moduleConstants, ShuffleboardTab swerveTab) {
...
+    
   }
 
   private void configCANcoder() {
     m_CANcoder.configFactoryDefault();
     m_CANcoder.configAbsoluteSensorRange(AbsoluteSensorRange.Unsigned_0_to_360);
     m_CANcoder.configSensorDirection(DriveConstants.kModuleConstants.canCoderInvert);
     m_CANcoder.configSensorInitializationStrategy(SensorInitializationStrategy.BootToAbsolutePosition);
-    m_CANcoder.configFeedbackCoefficient(0.087890625, "deg", SensorTimeBase.PerSecond);
+    m_CANcoder.configFeedbackCoefficient(360 / 4096, "deg", SensorTimeBase.PerSecond);
```
AnthonyFurman, line 78: needs to be doubles or else int divison

src/main/java/frc/robot/subsystems/Drivetrain.java
PandaNinja10: added where the yaw is from in java doc.

AnthonyFurman: I mean instead of doing getPose().getRotation() you can just do getYaw() which is probably more clear anyway

src/main/java/frc/robot/subsystems/Drivetrain.java
```diff
@@ -380,24 +246,28 @@ public void updateOdometry() {
...
-          }
+          if (currentTagPose == null) continue;
+
           Translation2d currentTagPoseTranslation = currentTagPose.toPose2d().getTranslation();
-          
-          // If the current april tag position is closer than the closest one, this makes makes it the closest
-          if (j == 0 || currentEstimatedPoseTranslation.getDistance(currentTagPoseTranslation) < currentEstimatedPoseTranslation.getDistance(closestTagPoseTranslation)) {
+
+          // If the current april tag position is closer than the closest one, this makes it the closest
+          if (j == 0 ||
```
AnthonyFurman, line 268: closestTagPoseTranslation starts off as (0, 0) then when the first index happens is set to that (by checking if j==0) but before that there is `if (currentTagPose == null) continue;` so theoretically if the first one is null and the pose is closer to (0,0) there could be issues. I would suggest making closestTagPoseTranslation = null; and then instead of j == 0, do closestTagPoseTranslation == null.

PandaNinja10: this ijust supose ot be formating maybe add this to the grid alignment PR

src/main/java/frc/robot/constants/swerve/DriveConstants.java
richtan, line 1: I don't think adding constants that we will never use is necessary.

PR #87, opened 2023-03-17
Set Cone as default to avoid fault

GLRoylance:
## Overview

This PR fixes a bug where access to Shuffleboard retuns null instead of a Cone or Cube; the null subsequently causes a fault.

## Tests Ran

SimGUI blew up during auto; with fix some auto routines run.


AnthonyFurman:
cone setter was removed

PR #94, opened 2023-03-25
drivetrain sim updates DrawMechanism's dist to grid 

MagicEmu:
## Overview

This PR will do what title says


src/main/java/frc/robot/subsystems/Drivetrain.java
```diff
@@ -370,7 +373,7 @@ public void resetModulesToAbsolute() {
   public void updateOdometry() {
     // Updates pose based on encoders and gyro. NOTE: must use yaw directly from gyro!
     m_poseEstimator.update(Rotation2d.fromDegrees(m_pigeon.getYaw()), getModulePositions());
-
+    m_mechanism.setDistanceToGrid(Math.max(m_poseEstimator.getEstimatedPosition().getX(), 0));
```
AnthonyFurman, line 376: this would set the grid distance to zero when the robot X pose is zero, but (0, 0) in the coordinate system is actually behind the grids.
May be clearer in this image, the very left is zero X, but the edge of the grids is further. In this case the robot X is 0.6m but it's on top of the girds still.
image
I would find the distance to the edge of the grids, then add half the robot length. The robot length is already a constant in DriveConstants, kRobotWidthWithBumpers. If you look at Node.java line 58, you can see how it gets the distance to the node tape. Note it's different for red/blue alliance.

PandaNinja10 (approved):

PR #95, opened 2023-03-25
Perpetual path planner command

PandaNinja10:
adds perpetual option for pathplanner command to prevent the command from stopping


src/main/java/frc/robot/commands/auto/PathPlannerCommand.java
```diff
@@ -23,6 +23,7 @@
 public class PathPlannerCommand extends SequentialCommandGroup {
   
   Alliance alliance;
+  boolean m_perpetual;
```
AnthonyFurman, line 26: does this need to be a member variable? Probably alliance can be removed as well

src/main/java/frc/robot/commands/auto/PathPlannerCommand.java
```diff
@@ -54,16 +60,29 @@ public PathPlannerCommand(List<PathPlannerTrajectory> pathGroup, int pathIndex,
...
+        pathGroup.get(pathIndex), 
+        () -> Conversions.absolutePoseToPathPlannerPose(
+        drive.getPose(), DriverStation.getAlliance()), // Pose supplier
+        drive.getPathplannerXController(), // X controller can't normal PID as pathplanner has Feed Forward 
+        drive.getPathplannerYController(), // Y controller can't normal PID as pathplanner has Feed Forward 
+        drive.getPathplannerRotationController(), // Rotation controller can't normal PID as pathplanner has Feed Forward 
+        (chassisSpeeds) -> { drive.setChassisSpeeds(chassisSpeeds, false); }, // chassis Speeds consumer
+        true,  // use Alliance color
+        drive // Requires this drive subsystem
+      )
```
AnthonyFurman, line 85: I feel like this could be cleaner. Maybe make a static function in PPSwerveControllerCommandPerpetual that returns either a the perpetual version or the normal version based on a bool (along with taking all the other params). that prevents errors in case someone modifies something in one of the parameters but forgets to in the other parameter list.
so it would be like
PathPlannerCommand.java:
```
...
addCommands(
 new InstantCommand(...), ...
 PPSwerveControllerCommandPerpetual.create(path, pose, .... , isPerpetual)
);
```
PPSwerveControllerCommandPerpetual.java
```
public static PPSwerveControllerCommand create(Path path, pose pose, ... boolean isPerpetual) {
   if (isPerpetual) {
        // the perpetual version
        return new PPSwerveControllerCommandPerpetual(path, pose, ...);
   } else {
        // the normal version
       return new PPSwerveControllerCommand(path, pose, ...);
  }
}
```

PandaNinja10: I think I did what you were seggusting. But I am still not very sure exactly what you ment.

AnthonyFurman: That's not what I was suggesting. You need a static function in PPSwerveControllerCommandPerpetual like the one I wrote. I can talk to you about it in class

AnthonyFurman (approved):

src/main/java/frc/robot/commands/auto/PathPlannerCommand.java
AnthonyFurman: Not exactly. I wanted all the parameters that might change to be inputted in one place. I did that

AnthonyFurman (approved):

PR #101, opened 2023-03-29
Supplier command

PandaNinja10:
Adds Commad used to reslove potental issues useing the Proxy Command. The command runs as if it were to command from the supplier upon initalization.


AnthonyFurman:
fix spaces and test in sim. Then this will be approved.

PandaNinja10:
Tested in Sim no visable issues

AnthonyFurman (approved):

PR #102, opened 2023-03-30
Wristclean

GLRoylance:
## Overview

This PR cleans up the wrist. Changes some physical values so simulation can lift the wrist. Using internal scale and offset of DutyCycleEncoder. Use DutyCycleEncoderSim so tests will exercise isReal() code rather than special simulation code. Make some methods private. Make some private fields default/package-private so unit tests can use them. Add unit tests that include simulating wrist movement over 1.6 seconds. Investigate PID and feedforward values.

## Design Doc

Programming practices.

## Tests Ran

Unit tests. SimGUI now runs autonomous routines (before routines would wait for wrist movement that never finished).

Believe DutyCycleEncoder.setPositionOffset(WristConstants.kEncoderOffset) does the right thing, but that has not been tested.


src/main/java/frc/robot/subsystems/Wrist.java
```diff
@@ -55,20 +43,46 @@ public Wrist(ShuffleboardTab wristTab) {
...
 
-    m_wristTab = wristTab;
-
     // configure the encoder
     m_absEncoder = new DutyCycleEncoder(WristConstants.kAbsEncoderPort);
+    // Cleaner encoder implementation
+    // offset to zero (arm horizontal)
+    m_absEncoder.setPositionOffset(WristConstants.kEncoderOffset);
+    // scale to radians and invert direction
+    m_absEncoder.setDistancePerRotation(-2.0 * Math.PI);
```
AnthonyFurman, line 52: We should test that these work, I think we had some issues setting the factors like this. On the other hand that might be because we were using getAbsolutePosition() and not getDistance()

src/main/java/frc/robot/subsystems/Wrist.java
```diff
@@ -129,35 +152,59 @@ public void setEnabled(boolean enable) {
    * @return the absolute encoder position in rotations, zero being facing forward
    */
   public double getAbsEncoderPos() {
-    if (RobotBase.isSimulation()) return m_armSim.getAngleRads();
+    // This is the wrong way to simulate the encoder. Should just use m_absEncoder.getDistance()
+    // if (RobotBase.isSimulation()) return m_armSim.getAngleRads();
     // inverted to make rotating towards stow positive
     // offset makes flat, facing out, zero
-    return (-m_absEncoder.getAbsolutePosition() + WristConstants.kEncoderOffset) * 2 * Math.PI; 
+    // return (-m_absEncoder.getAbsolutePosition() + WristConstants.kEncoderOffset) * 2 * Math.PI;
```
AnthonyFurman, line 159: the reason for that was that we thought the encoderSim was broken. We can go over it with Michael C. who spent a while trying to get it to work and eventually just did that

AnthonyFurman:
Now testWristMovement() is a unit test I can get behind.

Maybe we should go over these fixes with some of the students so they know the difference between getAbsolutePosition() and getDistance(), as well as the encoder simulation fix. They can remove the irrelevant comments

We noticed the MOI was too high, I think they calculated it wrong, probably wrong units. 

Are you fine with making the encoder and pid more public?

src/main/java/frc/robot/subsystems/Wrist.java
GLRoylance: DutyCycleEncoderSim.setDistance() is broken. There's a bug report and a fix. It should be fixed in the next release. A workaround is used.

GLRoylance:
A few days ago I pointed out that kLength (arm length) was interpreted as 16 meters rather than 16 inches. I do not expect most programmers to have an intuition about MOI.

Generally, good practice wants to restrict access as much as possible, so private is preferred.

IIRC, Java wants getters because methods can override but fields cannot. Consequently, interfaces use getters rather than fields. In a non-library context, there would be little difference between a getter and a protected/package-private field. There are camps that advocate both ways.

In this situation, I wanted access for the unit tests (within package) and did not want to give access outside the package. The PID was used in some now-commented-out print methods, so m_pid and m_pidPower could return to private. Other tests need access to the encoder and encoderSim to test them.

We have not had any code review meetings. Think our design review sessions but with the programmers reviewing code on the screen. Such sessions would cover coding practices such as replacing sequential ifs with switch, employing type restrictions (e.g., the whole series of WristConstants.kStow should be an enum), and keeping things simple (e.g., the Wrist is never disabled).

GLRoylance:
Unit tests should be used. Most unit tests look trivial or nonsensical, but think about it. Such code proves that we know how to do something. How did I find out that DutyCycleEncoderSim.setDistance() did not work? I tried some obvious code that should have worked but did not.

I had used some print methods to watch the PID loop (it was not moving despite motor powers much greater than one), and those print methods needed access to some Wrist internals. I have removed those PID print methods and returned the field access to restricted.

More generally, I have not commented out unit tests. I have commented out some asserts that should not have succeeded. I have disabled two failing unit tests because DutyCycleEncoderSim is broken.

Yes, I want the code merged in its current state. Merging does not require perfection. Code can change.

The comment you cited is the rationale for changing a public method to private. Consider it an educational note. Methods that are used locally do not need to be public and probably should not be public.

I want all the programmers to review all the code -- not just this PR. Quality and education are benefits of code reviews.

The format is simple. What is the code supposed to do? How does it do it? The group then goes through the code and critiques it. Are there better or simpler ways of coding? Does the code do what it is supposed to do? Are there circumstances that the code does not consider?

The team is weak in simulation and testing.

AnthonyFurman:
Maybe I was unclear, I do support these unit tests that are actually testing functionality. I don't support unit tests that just check 5 == 5.
Also, we knew that the encoderSim wasn't working without unit tests, because it wasn't working.

Is there a reason to keep the commented out section there? 
Is there a reason to add unit tests that are disabled? We will have to make a PR anyway when WPI updates, so I would add it in that PR.

What I mean about the comments is they are not actual "comments" explaining the code but notes you want the students to review. If I am trying to understand what the code does, I read the comments. Reading "Cleaner encoder implementation" or "changed to private method; not called from outside" would just make me more confused.
To me it seems more like an "educational note" as you call it, but the way we should have these educational notes is by going over them with students, not by merging them with main.
We can go over it with the students in a class soon.

Yes, that is theoretically what the programmers did when I had them make the subsystem review slideshow. Now not all the programmers are very experienced so they likely missed things. But for example Yichen did suggest the same thing you did about removing the manual m_enabled in wrist. Michael Kersey made a drivetrain cleaning PR. That's not enough yet in my opinion though. I'm supporting more code reviews. If you have a specific format in mind (i.e. number of people, how they present their findings, etc.) let me know.

AnthonyFurman:
@GLRoylance It seems like testWristMovement() works then doesn't work. I'll build once, then it works, I'll build again, and it fails. Without changing any code in between builds.

Also, should we remove the disabled tests?

src/main/java/frc/robot/subsystems/Wrist.java
```diff
@@ -133,41 +159,59 @@ public void setEnabled(boolean enable) {
    * @return the absolute encoder position in rotations, zero being facing forward
    */
   public double getAbsEncoderPos() {
-    // inverted to make rotating towards stow positive
-    // offset makes flat, facing out, zero
-    double pos = (-m_absEncoder.getAbsolutePosition() + WristConstants.kEncoderOffset) * 2 * Math.PI; 
-    if (pos > WristConstants.kMaxPos) {
+    double pos = m_absEncoder.getDistance();
+    if (pos > WristConstants.kMaxPos || pos > WristConstants.kMaxPos) {
```
PandaNinja10, line 163: I think this is supose to be cheking the max & min not the max twice

AnthonyFurman (approved):

PR #105, opened 2023-04-02
Proposal to improve balance command

vfurman-gh:
## Overview

Improve balance command to start stopping as soon as we are reasonably certain the charge station angle has changed. 
To avoid measurement noise triggering undesired stopping and yet be able to detect the change early, we require that the angle has changed by more than X  degrees since the initialization of the command. Actual values should be tuned by by testing on robot.

## Design Doc

No design doc, please try if it works on actual robot.

## Tests Ran

Compiles, but no actual tests were run.


AnthonyFurman:
Usually we still want the robot to continue driving forward even after it is on the charge station. 
If anything, it should start stopping when the charge station becomes close to flat, or right before that.

vfurman-gh:
This PR makes the robot to continue moving on the charge station until it observes that the station starts to level up (pitch decreases by 2 degrees since the beginning of the command). So you probably misunderstood the proposal.

AnthonyFurman:
We talked and I suspect the robot goes up onto the charge station, it's wheels start touching the ground, thus it's angle increases at first. So it should compare to the max angle and not the starting angle.

AnthonyFurman (approved):

PR #110, opened 2023-04-11
Set wrist voltage compensation saturation

richtan:
This PR sets the voltage compensation saturation for the wrist to 12 volts because it is required for voltage compensation to work. See [docs](https://store.ctr-electronics.com/content/api/java/html/interfacecom_1_1ctre_1_1phoenix_1_1motorcontrol_1_1_i_motor_controller.html#ad8eb8bd875f79543bfc56ddf191a4220)

### DIFF
--- a/src/main/java/frc/robot/subsystems/Wrist.java
+++ b/src/main/java/frc/robot/subsystems/Wrist.java
@@ -45,6 +45,8 @@ public class Wrist extends SubsystemBase {
       WristConstants.kPeakCurrentDuration);
     m_motor.setNeutralMode(WristConstants.kNeutralMode);
     m_motor.setInverted(WristConstants.kMotorInvert); 
+
+    m_motor.configVoltageCompSaturation(Constants.kRobotVoltage);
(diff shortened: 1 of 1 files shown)


AnthonyFurman:
The motor is created using MotorFactory which sets that value.

richtan:
I am not sure why this configuration is split between two files. This would be better if both lines were either in motor factory or both in wrist.java

AnthonyFurman:
well theoretically if you want to enable/disable voltage compensation that makes it easier and affects nothing.
If you want you can make it enable the voltage compensation in MotorFactory. The reason we had it disabled is so that we could still use setVoltage on some motors (I think we were using it for characterization). That's not used anymore and can be done differently anyway so it can be enabled

AnthonyFurman:
To clarify, if you set voltage compensation enabled and then use setVoltage CTRE throws a runtime error and the robot code breaks. So I would double check which motors are and aren't using voltage compensation if you decide to change MotorFactory

